PR #5, opened 2020-02-15
Update run-full-node-mainnet.md

moonstash:
Added further clarification -- Those new to linux wishing to work through this are still confused, some appear to think the block of text was one large command as opposed to being various commands. Clarified the need to port forward to make that requirement more obvious.


assafmo:
Thanks @moonstash!

PR #16, opened 2020-02-25
update README.md

moonstash:
typo


TomL94:
Thanks! Can you also rename the file itself so the link will not break?

assafmo:
https://github.com/enigmampc/EnigmaBlockchain/commit/5464339a3483466e219521ff0d20376d4020d37c

PR #122, opened 2020-03-24
replacing the wasmer runtime with a wasmi runtime running inside an SGX enclave

reuvenpo:
EDIT: This is still a WIP.

This PR includes the following pieces to it, from top to bottom:
- [x] Patch `go-cosmwasm` to use local dependencies
- [x] Copy `cosmwasm-vm` to a new directory containing the new wasmi+sgx code
- [x] Adjust `cosmwasm-sgx-vm` to call the enclave (`wasmi-runtime`) instead of using `wasmer` #129 
- [x] Define the interface between `cosmwasm-sgx-vm` and the `wasmi-runtime`
    * The types in this interface are defined in `cosmwasm/lib/enclave-ffi-types` and have autogenerated C bindings. See the `.edl` file in `cosmwasm/lib/wasmi-runtime` to see how they are used.
- [x] Implement safe wrappers around the FFI on the `cosmwasm-sgx-vm` side
- [x] Write the `.edl` file describing the interface between the enclave and the outside world
- [x] Implement safe wrappers around the FFI on the `wasmi-runtime` side
- [x] Use `wasmi` to execute the wasm contracts in the environment they expect, as defined in `cosmwasm` #123


README.md
```diff
@@ -31,8 +31,8 @@ Mainnet is out! Get the latest release at https://github.com/enigmampc/EnigmaBlo
 # Implementation Discussions
 
 - [Secret Contracts on Enigma Blockchain](https://forum.enigma.co/t/secret-contracts-on-enigma-blockchain/1284)
-- [Input/Output/State Encryption/Decryption protocol](https://forum.enigma.co/t/input-output-state-encryption-decryption-protocol/1325)
 - [Network key management/agreement](https://forum.enigma.co/t/network-key-management-agreement/1324)
+- [Input/Output/State Encryption/Decryption protocol](https://forum.enigma.co/t/input-output-state-encryption-decryption-protocol/1325)
```
assafmo, line 35: Should also be applied to master

PR #123, opened 2020-03-26
Wasmi enclave

assafmo:


cosmwasm/lib/wasmi-runtime/Enclave_private.pem
```diff
@@ -0,0 +1,39 @@
+-----BEGIN RSA PRIVATE KEY-----
```
toml01: Whoopsy doopsy

cosmwasm/lib/Makefile
```diff
@@ -0,0 +1,172 @@
...
+
+######## EDL Settings ########
+
+Enclave_EDL_Files := ./wasmi-runtime/Enclave_t.c ./wasmi-runtime/Enclave_t.h ./sgx-vm/Enclave_u.c ./sgx-vm/Enclave_u.h
+
+######## APP Settings ########
+
+App_Rust_Flags := --release
+App_SRC_Files := $(shell find sgx-vm/ -type f -name '*.rs') $(shell find sgx-vm/ -type f -name 'Cargo.toml')
+App_Include_Paths := -I ./sgx-vm -I./include -I$(SGX_SDK)/include -I$(CUSTOM_EDL_PATH)
```
reuvenpo: Isn't the space after the fist `-I` not supposed to be there?

cosmwasm/lib/Makefile
```diff
@@ -0,0 +1,172 @@
...
+.PHONY: all
+all: enclave-ffi-types $(App_Name) $(Signed_RustEnclave_Name)
+
+.PHONY: enclave-ffi-types
+enclave-ffi-types:
+	cd enclave-ffi-types && cargo build
+	mkdir -p ./sgx-vm/target/headers
+	mkdir -p ./wasmi-runtime/target/headers
+	cp -f ./enclave-ffi-types/target/headers/enclave-ffi-types.h ./wasmi-runtime/target/headers
+	cp -f ./enclave-ffi-types/target/headers/enclave-ffi-types.h ./sgx-vm/target/headers
```
reuvenpo: All of this already happens as part of `enclave-ffi-types`'s build script, and you don't need to do it at all :) Any package that depends on `enclave-ffi-types` will have these files available when building, before it gets built itself

cosmwasm/lib/Makefile
```diff
@@ -0,0 +1,172 @@
...
+	Trts_Library_Name := sgx_trts
+	Service_Library_Name := sgx_tservice
+endif
+Crypto_Library_Name := sgx_tcrypto
+KeyExchange_Library_Name := sgx_tkey_exchange
+ProtectedFs_Library_Name := sgx_tprotected_fs
+
+RustEnclave_C_Files := $(wildcard ./wasmi-runtime/*.c)
+RustEnclave_C_Objects := $(RustEnclave_C_Files:.c=.o)
+RustEnclave_Include_Paths := -I$(CUSTOM_COMMON_PATH)/inc -I$(CUSTOM_EDL_PATH) -I$(SGX_SDK)/include -I$(SGX_SDK)/include/tlibc -I$(SGX_SDK)/include/stlport -I$(SGX_SDK)/include/epid -I ./wasmi-runtime -I./include
```
reuvenpo: Near the end of the line, you probably have a redundant space in `-I ./wasmi-runtime`

cosmwasm/lib/wasmi-runtime/Enclave_private.pem
assafmo:
```suggestion
-----BEGIN RSA PRIVATE KEY-----asdasd
```

assafmo: This is fine, it's just a test file from intel for now (google its content)

assafmo: Taken from here: https://github.com/intel/intel-sgx-ssl/blob/master/Linux/sgx/test_app/enclave/TestEnclave_private.pem

cosmwasm/lib/wasmi-runtime/Cargo.toml
```diff
@@ -5,7 +5,43 @@ authors = ["Reuven Podmazo <reuven@enigma.co>"]
...
+sgx_tprotected_fs = { path = "../incubator-teaclave-sgx-sdk/sgx_tprotected_fs" }
+sgx_trts = { path = "../incubator-teaclave-sgx-sdk/sgx_trts" }
+sgx_tse = { path = "../incubator-teaclave-sgx-sdk/sgx_tse" }
+sgx_tseal = { path = "../incubator-teaclave-sgx-sdk/sgx_tseal" }
+sgx_tstd = { path = "../incubator-teaclave-sgx-sdk/sgx_tstd" }
+sgx_tunittest = { path = "../incubator-teaclave-sgx-sdk/sgx_tunittest" }
+sgx_types = { path = "../incubator-teaclave-sgx-sdk/sgx_types" }
+sgx_ucrypto = { path = "../incubator-teaclave-sgx-sdk/sgx_ucrypto" }
+sgx_unwind = { path = "../incubator-teaclave-sgx-sdk/sgx_unwind" }
+sgx_urts = { path = "../incubator-teaclave-sgx-sdk/sgx_urts" }
```
reuvenpo: 1. The first section there says you're only pulling these dependencies when NOT building for SGX.
2. You then override the github links with local paths.
3. You should be pulling only the dependencies you need, and you're not using most of these
4. All of these are available on crates.io. please just specify normal crates.io dependencies.

cosmwasm/lib/.gitignore
```diff
@@ -0,0 +1,9 @@
+lib*.a
+Cargo.lock
+Enclave_t.c
+Enclave_t.h
+Enclave_t.o
+
+Enclave_u.c
+Enclave_u.h
+Enclave_u.o
```
reuvenpo: Please use more specific .gitignore files in each directory. not each directory has all these files, and `wasmi-runtime` should definitely keep its `Cargo.lock` file.

cosmwasm/lib/Makefile
```diff
@@ -0,0 +1,172 @@
...
+
+$(App_Enclave_u_Object): sgx-vm/Enclave_u.o
+	$(AR) rcsD $@ $^
+	cp $(App_Enclave_u_Object) ./lib
+
+$(App_Name): enclave-ffi-types $(App_Enclave_u_Object) $(App_SRC_Files)
+	@cd sgx-vm && SGX_SDK=$(SGX_SDK) cargo build $(App_Rust_Flags)
+	@echo "Cargo  =>  $@"
+	mkdir -p bin
+	cp $(App_Rust_Path)/sgx-vm ./bin
```
reuvenpo: The rules that run cargo here should run as their own rule before `$(App_Enclave_u_Object)`, and instead of the `enclave-ffi-types` rule

cosmwasm/lib/wasmi-runtime/Enclave.config.xml
```diff
@@ -0,0 +1,12 @@
+<!-- Please refer to User's Guide for the explanation of each field -->
+<EnclaveConfiguration>
+  <ProdID>0</ProdID>
```
reuvenpo: in enigma-core the ProdID was 1. you should check what it means

cosmwasm/lib/wasmi-runtime/Enclave_private.pem
reuvenpo: If it's not doing anything we should remove this file from the PR

cosmwasm/lib/wasmi-runtime/Xargo.toml
```diff
@@ -0,0 +1,91 @@
+[dependencies]
+alloc = {}
+
+[dependencies.sgx_types]
+path = "../incubator-teaclave-sgx-sdk/sgx_types"
+stage = 1
```
reuvenpo: Similarly to my comment above, this is n=probably unneeded.
also, you can rewrite this block as the following line, under the `[dependencies]` section:
```
sgx_types = { path = "../incubator-teaclave-sgx-sdk/sgx_types", stage = 1 }
```
and the other blocks below similarly

cosmwasm/lib/Makefile
```diff
@@ -0,0 +1,172 @@
...
+wasmi-runtime/Enclave_t.o: $(Enclave_EDL_Files)
+	@$(CC) $(RustEnclave_Compile_Flags) -c wasmi-runtime/Enclave_t.c -o $@
+	@echo "CC   <=  $<"
+
+$(RustEnclave_Name): enclave compiler-rt wasmi-runtime/Enclave_t.o
+	cp incubator-teaclave-sgx-sdk/compiler-rt/libcompiler-rt-patch.a ./lib
+	@$(CXX) wasmi-runtime/Enclave_t.o -o $@ $(RustEnclave_Link_Flags)
+	@echo "LINK =>  $@"
+
+$(Signed_RustEnclave_Name): enclave-ffi-types $(RustEnclave_Name)
```
reuvenpo: I believe that you don't need the `enclave-ffi-types` rule here at all, since the `$(RustEnclave_Name)` rule runs tthe `enclave` rule, which runs cargo/xargo before running anything else, and that should place the headers in their right place.

cosmwasm/lib/Makefile
assafmo: Both packages depend on `enclave-ffi-types` and they don't have these files, so...

assafmo: It doesn't matter

cosmwasm/lib/Makefile
assafmo: It doesn't matter

cosmwasm/lib/Makefile
assafmo: It doesn't matter

cosmwasm/lib/wasmi-runtime/Cargo.toml
assafmo: We only copied this from the sample project, trying to make it work rn

cosmwasm/lib/wasmi-runtime/Enclave_private.pem
toml01: @assafmo I know, jk :)

cosmwasm/lib/Makefile
assafmo: Yep but we get an error about missing `enclave-ffi-types.h` in those directories so we added this rule for now

reuvenpo: can you please show me the error when building only this rule?

cosmwasm/lib/Makefile
```diff
@@ -0,0 +1,166 @@
...
+
+sgx-vm/Enclave_u.o: $(Enclave_EDL_Files)
+	@$(CC) $(App_C_Flags) -c sgx-vm/Enclave_u.c -o $@
+	@echo "CC   <=  $<"
+
+$(App_Enclave_u_Object): sgx-vm/Enclave_u.o
+	$(AR) rcsD $@ $^
+	cp $(App_Enclave_u_Object) ./lib
+
+foo:
```
reuvenpo: Please rename this :smile:

cosmwasm/lib/Makefile
reuvenpo: resolved in call

cosmwasm/lib/wasmi-runtime/Cargo.toml
reuvenpo: resolved in call

cosmwasm/lib/Makefile
reuvenpo: resolved in call

cosmwasm/lib/Makefile
```diff
@@ -0,0 +1,166 @@
...
+wasmi-runtime/Enclave_t.o: $(Enclave_EDL_Files)
+	@$(CC) $(RustEnclave_Compile_Flags) -c wasmi-runtime/Enclave_t.c -o $@
+	@echo "CC   <=  $<"
+
+$(RustEnclave_Name): enclave compiler-rt wasmi-runtime/Enclave_t.o
+	cp incubator-teaclave-sgx-sdk/compiler-rt/libcompiler-rt-patch.a ./lib
+	@$(CXX) wasmi-runtime/Enclave_t.o -o $@ $(RustEnclave_Link_Flags)
+	@echo "LINK =>  $@"
+
+$(Signed_RustEnclave_Name): enclave-ffi-types $(RustEnclave_Name)
```
reuvenpo: remove the reference to `enclave-ffi-types`

cosmwasm/lib/Makefile
```diff
@@ -0,0 +1,166 @@
...
+.PHONY: compiler-rt
+compiler-rt:
+	$(MAKE) -C incubator-teaclave-sgx-sdk/compiler-rt/ 2> /dev/null
+
+.PHONY: clean
+clean:
+	@rm -f $(App_Name) $(RustEnclave_Name) $(Signed_RustEnclave_Name) wasmi-runtime/*_t.* sgx-vm/*_u.* lib/*.a
+	@cd wasmi-runtime && cargo clean && rm -f Cargo.lock
+	@cd sgx-vm && cargo clean && rm -f Cargo.lock
+	@cd enclave-ffi-types && cargo clean && rm -f Cargo.lock
```
reuvenpo: remove this line

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/cosmos/cosmos-sdk v0.38.1
 	github.com/gorilla/mux v1.7.3
+	github.com/pkg/errors v0.9.1
```
reuvenpo: why did you add this?

cosmwasm/lib/wasmi-runtime/Xargo.toml
reuvenpo: done

cosmwasm/lib/wasmi-runtime/Enclave.config.xml
assafmo: Screenshot from 2020-04-08 00-40-51
Screenshot from 2020-04-08 00-40-59

Source: [Intel_SGX_Developer_Reference_Linux_2.6_Open_Source.pdf](https://github.com/enigmampc/EnigmaBlockchain/files/4447128/Intel_SGX_Developer_Reference_Linux_2.6_Open_Source.pdf)

assafmo: Downloaded the pdf from https://download.01.org/intel-sgx/linux-2.6/docs/Intel_SGX_Developer_Reference_Linux_2.6_Open_Source.pdf

cosmwasm/lib/wasmi-runtime/Enclave_private.pem
assafmo: It's used to sign the enclave so we do need this file: https://github.com/enigmampc/EnigmaBlockchain/blob/6cd448d376da812de940b7f20c870ad64f52c398/cosmwasm/lib/Makefile#L149

assafmo: But we'll want to use our own private key in production. This is fine for now.

go.mod
assafmo: `go mod` discovered it in here:   https://github.com/enigmampc/EnigmaBlockchain/blob/6cd448d376da812de940b7f20c870ad64f52c398/cosmwasm/lib/incubator-teaclave-sgx-sdk/samplecode/ue-ra/ue-ra-client-go/cert.go#L10

So my IDE auto-added it to `go.mod`

cosmwasm/lib/wasmi-runtime/Enclave.config.xml
assafmo: It's our product number... We also have an ISV number from Intel we should put in this file.

assafmo: In core it was 1 and in km it was 2 https://github.com/enigmampc/enigma-core/blob/3c04c2625741daa693cdc1a867db03bc1dc3ead2/enigma-principal/enclave/Enclave.config.xml#L3

It's just our product number. We also have an ISV number from Intel that we should put there.

PR #129, opened 2020-03-30
Changes to sgx vm

reuvenpo:
This PR implements all the changes in sgx-vm and go-cosmwasm required to use the enclave correctly, instead of wasmer.


reuvenpo:
I'm closing this PR since it was merged into #123, which will be later merged into #122, which is the branch that this PR was targeting.

PR #134, opened 2020-04-08
Docker build for running enigma on MacOS for dev  (#1)

levackt:
I couldn't get a dev environment working on Mac though I see the makefile has build_macos, not sure if install is meant to work already, but I did it get running using this docker build.

I don't know if its useful for anyone else, or can be replaced by docs and scripts using the existing docker build instead.


assafmo:
Doesn't that https://github.com/enigmampc/EnigmaBlockchain/blob/master/Dockerfile work for you?

PR #257, opened 2020-06-16
Set up CI with Azure Pipelines

Cashmaney:


cankisagun:
@Cashmaney does this have anything to do with our discussion on launching a secret node on azure marketplace?

Cashmaney:
@cankisagun This specifically is just to be able to run automatic tests on SGX nodes, but on the marketplace side I registered Enigma on the Partner Portal (waiting for approval) so we will be able to move forward with that soon

cankisagun:
If you have a receipt or something let me know (email me) and I'll pass it to the microsoft team to get us approved faster

Cashmaney:
Should be pretty standard stuff - I don't expect it to take more than a few hours, but I'll let you know if we need some extra firepower. Thanks.

PR #289, opened 2020-06-19
adding new attack.

FishmanL:


assafmo:
Related thread: https://chat.scrt.network/channel/general?msg=R4tGADnpE8R2TredS

docs/protocol/encryption-specs.md
```diff
@@ -615,6 +615,18 @@ For example, An original contract with a permissioned getter, such that only whi
...
+
+Specifically, assume for millionaire's that you have a contract where one person inputs their amount of money, then the other person does, then the contract sends them both a single bit saying who's is more -- this is the simplest implementation for Millionaire's problem solving, and it reveals the minimum amount of information (a single bit). As person 2, binary search the interval of possible money amounts person 1 could have -- say you know person 1 has less than 1 trillion dollars. first, query with 500 billion with your node detached from the wider network, get the single bit out (whether the true value is higher or lower), then repeat
+in log(n) tries (where n is the size of the interval) you'll have the exact value of person 1's input.
+
+The naive solution to this is requiring the node to successfully send out the answer to person 1 and broadcast the erasure of the data of person 1 and person 2 in the DHT before revealing an answer, but even that's imperfect since you can reload the contract and replay the network state up to that broadcast, then perform the attack with repeated rollbacks.
```
assafmo, line 625: Can you better explain this paragraph?

docs/protocol/encryption-specs.md
```diff
@@ -615,6 +615,18 @@ For example, An original contract with a permissioned getter, such that only whi
...
 
+## More Advanced Tx Replay attacks -- search to decision for Millionaire's problem
+
+This attack provides a specific example of a TX replay attack extracting the full information of a client based on replaying a TX.
+
+Specifically, assume for millionaire's that you have a contract where one person inputs their amount of money, then the other person does, then the contract sends them both a single bit saying who's is more -- this is the simplest implementation for Millionaire's problem solving, and it reveals the minimum amount of information (a single bit). As person 2, binary search the interval of possible money amounts person 1 could have -- say you know person 1 has less than 1 trillion dollars. first, query with 500 billion with your node detached from the wider network, get the single bit out (whether the true value is higher or lower), then repeat
```
assafmo, line 622: This can be explained better. How can the 2nd player query the contract log(n) times? Taking the node offline is not enough. Also the numbers (1 trillion etc) are irrelevant.

guyz:
Adding my comment from the Secret Network chat:

I don't see this as a viable attack vector. Secure computation does not tackle the problem of output privacy, and thus is not part of the threat model.

Each contract developer will have to weigh the affect of output leakage in their contract properly.

Surely, in the future, based on use cases and needs, we will consider developing a DP extension/privacy budget to mitigate this, or some network synch mechanism to prevent replays altogether (these are technically, separate issues, but in this PR they've been merged together). For now, a contract developer, for example, may decide that a big enough anonymity set is required in a given application, and N=2 (e.g., in the classical millionare's problem) doesn't count.

The bounties for attacks are more around protocol-level vulnerabilities for the given protocol, and even there - a PoC for attacks would generally be required (even if it's a theoretical attack, we need some proof that it works)

PR #262, opened 2020-06-18
Cosmwasm 0.9

reuvenpo:
This PR, in isolation, is too big to comprehend. What follows is an explanation of how i assemebled its contents:

## Background
The `cosmwasm/`, `go-cosmwasm/`, and `x/compute` and `cmd/` directories in this repository ([enigmampc/SecretNetwork](https://github.com/enigmampc/SecretNetwork)) started as copies of the [CosmWasm/cosmwasm](https://github.com/CosmWasm/cosmwasm), [CosmWasm/go-cosmwasm](https://github.com/CosmWasm/go-cosmwasm), and [CosmWasm/wasmd](https://github.com/CosmWasm/wasmd) repositories respectively, at their `v0.7.0` release.
We have since then modified these repositories considerably, according to our custom needs, but in parallel to this, the CosmWasm team kept developing their own projects, fixing bugs, and adding features.
During this period, I (@reuvenpo) also contributed significant amounts of code to their repositories, and helped bring these components to a production-ready state.

## Merge Strategy
The time has now come to finally merge all of the work done on both sides back into a coherent version in our repository.
In order to do this, I cloned the three CosmWasm repositories under our namespace:
 [enigmampc/cosmwasm](https://github.com/enigmampc/cosmwasm), [enigmampc/go-cosmwasm](https://github.com/enigmampc/go-cosmwasm), and [enigmampc/wasmd](https://github.com/enigmampc/wasmd).

For the `go-cosmwasm` and `wasmd` repositories, I first checked out the `v0.7.0` tag, created a new branch called `enigma-master`, and then replaced the content there with the relevant content from commit 213e32c in this repository. This represented a squash of all the work we did into one commit. I then used the usual `git merge` tool to perform most of the work, and manually resolved the couple-dozed conflicts that did occur (I merged a commit a little after `0.9.0-alpha1`). Later I also tweaked a few lines where necessary because of slight API changes.

[...]

PR #307, opened 2020-06-24
Update enclave error handling

### LINKED ISSUE #279: Update gas usage tracking
>  track gas usage in the enclave just like cosmwsam does it (this also means return gas_used from some/all the imports.

### LINKED ISSUE #281: Update error handling for 0.9
Pass VmErrors and gas usage from imports back to the enclave and then back to the caller

reuvenpo:
closes #281 
also closes #279 (gas tracking in enclave)

also:
* removed dependency on wasmer (I forgot to do it on #262 originally)


cosmwasm/packages/enclave-ffi-types/src/types.rs
```diff
@@ -16,50 +16,19 @@ pub struct EnclaveBuffer {
 }
 
 impl EnclaveBuffer {
+    pub fn null() -> Self {
```
Cashmaney, line 19: Might be helpful to use this to impl `::default()`?

cosmwasm/packages/enclave-ffi-types/src/types.rs
```diff
@@ -105,86 +81,48 @@ pub enum EnclaveError {
...
-    // /// This error means that the random function had failed generating randomness.
-    // #[cfg(feature = "std")]
-    // RandomError {
-    //     err: rand::Error,
-    // },
-    // #[cfg(feature = "sgx")]
-    RandomError, // {
-                 //     err: sgx_types::sgx_status_t,
-                 // }
+#[display(fmt = "VmError")]
```
Cashmaney, line 92: Cool. Didn't know #[display] could be used like this

cosmwasm/packages/wasmi-runtime/src/crypto/aes_siv.rs
```diff
@@ -1,3 +1,4 @@
+use crate::crypto::errors::CryptoError;
```
Cashmaney, line 1: should be under the description

cosmwasm/packages/wasmi-runtime/src/crypto/aes_siv.rs
```diff
@@ -37,46 +37,28 @@ fn aes_siv_encrypt(
...
-        None => cipher.encrypt(&vec![[]], plaintext).map_err(|e| {
-            error!("aes_siv_encrypt error: {:?}", e);
-            CryptoError::EncryptionError
-        }),
-    }
+    let mut cipher = Aes128Siv::new(GenericArray::clone_from_slice(key));
+    cipher.encrypt(ad, plaintext).map_err(|e| {
+        error!("aes_siv_encrypt error: {:?}", e);
+        CryptoError::EncryptionError
+    })
```
Cashmaney, line 46: While I don't disagree with this change, try to avoid changing things that aren't directly related to your issue

cosmwasm/packages/wasmi-runtime/src/crypto/kdf.rs
```diff
@@ -67,8 +67,8 @@ pub mod tests {
     use super::{
         Keychain, CONSENSUS_SEED_SEALING_PATH, KEY_MANAGER, REGISTRATION_KEY_SEALING_PATH,
     };
+    use crate::crypto::errors::CryptoError;
```
Cashmaney, line 70: Might be cleaner to export it directly from crypto, so we can have a single import from crate::crypto

cosmwasm/packages/sgx-vm/src/wasmi/exports.rs
```diff
@@ -28,71 +27,127 @@ pub unsafe fn recover_buffer(ptr: UserSpaceBuffer) -> Option<Vec<u8>> {
 
 /// Read a key from the contracts key-value store.
 #[no_mangle]
-pub extern "C" fn ocall_read_db(context: Ctx, key: *const u8, key_len: usize) -> EnclaveBuffer {
+pub extern "C" fn ocall_read_db(
+    context: Ctx,
+    vm_error: *mut UntrustedVmError,
```
Cashmaney, line 32: Can't we just call this field `error`?

cosmwasm/packages/wasmi-runtime/src/wasm/db.rs
```diff
@@ -23,30 +23,30 @@ pub fn write_encrypted_key(
         scrambled_field_name
     );
 
-    let ad = derive_ad_for_field(&scrambled_field_name, &context)?;
+    let (ad, ad_used_gas) = derive_ad_for_field(&scrambled_field_name, &context)?;
```
Cashmaney, line 26: What's the reasoning for charging gas for this?

cosmwasm/packages/wasmi-runtime/src/wasm/db.rs
```diff
@@ -56,119 +56,167 @@ pub fn read_encrypted_key(
...
-            error!(
-                "read_db() got an error from ocall_read_db, stopping wasm: {:?}",
-                err
-            );
-            DbError::FailedRead
-        })?
+    match read_db(context, &scrambled_field_name) {
+        Ok((value, gas_used)) => match value {
+            Some(value) => match decrypt_key(&scrambled_field_name, &value, contract_key) {
+                Ok(encrypted) => Ok((Some(encrypted), gas_used)),
```
Cashmaney, line 62: `encrypted` -> `decrypted`?

cosmwasm/packages/wasmi-runtime/src/wasm/db.rs
```diff
@@ -56,119 +56,167 @@ pub fn read_encrypted_key(
...
 
 /// Safe wrapper around reads from the contract storage
-fn read_db(context: &Ctx, key: &[u8]) -> SgxResult<Option<Vec<u8>>> {
+fn read_db(context: &Ctx, key: &[u8]) -> Result<(Option<Vec<u8>>, u64), WasmEngineError> {
+    let mut ocall_return = OcallReturn::Success;
     let mut enclave_buffer = std::mem::MaybeUninit::<EnclaveBuffer>::uninit();
-    unsafe {
-        match imports::ocall_read_db(
-            enclave_buffer.as_mut_ptr(),
+    let mut vm_err = UntrustedVmError::null();
```
Cashmaney, line 105: Same as above, using `UntrustedVmError::default()` is probably cleaner.

cosmwasm/packages/sgx-vm/src/testing/instance.rs
```diff
@@ -91,7 +93,10 @@ impl Default for MockInstanceOptions<'_> {
             error_message: None,
 
             // instance
+            /*
             supported_features: features_from_csv("staking"),
+            */
+            supported_features: HashSet::new(),
             gas_limit: 500_000,
```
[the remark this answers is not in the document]
reuvenpo: גםמק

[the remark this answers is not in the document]
reuvenpo: done

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
```diff
@@ -110,6 +114,36 @@ impl ContractInstance {
         let ptr_to_region_in_wasm_vm = self.allocate(buffer.len() as u32)?;
         self.write_to_allocated_memory(buffer, ptr_to_region_in_wasm_vm)
     }
+
+    /// Track gas used inside wasmi
+    fn use_gas(&mut self, gas_amount: u64) -> Result<(), WasmEngineError> {
```
Cashmaney, line 119: At this point wouldn't we want to create a `GasTracker` or something to handle all these?

cosmwasm/packages/enclave-ffi-types/src/types.rs
reuvenpo: Yes, except it's shorter to write `EnclaveBuffer::null()` than `EnclaveBuffer::default()`, and it makes more sense in my head in this case. I can add it if you think it's more appropriate though.

cosmwasm/packages/enclave-ffi-types/src/types.rs
reuvenpo: neither did I :smile: I had to look it up.

reuvenpo: It is specific to the `derive-more` crate though, so it's not standard.

cosmwasm/packages/wasmi-runtime/src/crypto/aes_siv.rs
reuvenpo: :+1:

cosmwasm/packages/wasmi-runtime/src/crypto/kdf.rs
reuvenpo: good idea

cosmwasm/packages/wasmi-runtime/src/crypto/aes_siv.rs
reuvenpo: true, my bad

cosmwasm/packages/sgx-vm/src/wasmi/exports.rs
reuvenpo: it's specific for instances of the `VmError` type. (I'll explain what it is in a bigger comment below)

cosmwasm/packages/wasmi-runtime/src/wasm/db.rs
reuvenpo: the gas isn't charged by us. the gas is charged in cosmos because we access the storage when reading the previous key. this information is returned for tracking purposes.

cosmwasm/packages/wasmi-runtime/src/wasm/db.rs
reuvenpo: yes.

cosmwasm/packages/wasmi-runtime/src/wasm/db.rs
reuvenpo: as i said, i liked `null` better, but i can change it to default :)

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
reuvenpo: we definitely could. should be really simple, so if you want me to, just say "do it" :)

reuvenpo:
>Can you explain a little about what VmError and what UntrustedVmError are?

`VmError` is the standard error type for the `cosmwasm-sgx-vm` layer

>When do we need them?

During an ocall, we call into the original implementation of `db_read`, `db_write`, and `db_remove`. These call out all the way to the Go side. They return `VmError` when something goes wrong in this process. 

>What kind of errors are they passing, and why they are important?

This mostly involves errors around FFI (memory errors, format errors, panic, etc.) but most importantly it communicates cases where an out-of-gas event happens on the Go side, which needs to trigger finalization of the computation and unwind everything all the way back to the enclave, out of the enclave, and back to Go. Some of the type conversions in this PR forward this signal across the enclave boundaries. On the `cosmwasm-vm` side this is already implemented.

>Also, why did you choose to make them wrapper pointers instead of just an error code (ala sgx_status_t)?

`VmError` is not FFI safe, and isn't designed to be. Instead of reimplementing it, which would affect a ton of code, I just box it before returning from the ocall, and unbox it when returning from the ecall.

Specifically for the out-of-gas situation, it has two sources. One is in the cosmos SDK. In our codebase this is the instance of `VmError` which gets boxed and unboxed. The second is when we see that the sum of gas consumed inside and outside the enclave exceeds the allowed limit for the computation. in that case we signal a `WasmEngineError::OutOfGas` [which is converted to](https://github.com/enigmampc/SecretNetwork/blob/8806932eebb53e03dae995c08c0a6a398ce0f554/cosmwasm/packages/wasmi-runtime/src/wasm/errors.rs#L37) `EnclaveError::OutOfGas` [which is converted to](https://github.com/enigmampc/SecretNetwork/blob/8806932eebb53e03dae995c08c0a6a398ce0f554/cosmwasm/packages/sgx-vm/src/errors.rs#L348) `VmError::GasDepletion`

cosmwasm/packages/wasmi-runtime/src/wasm/db.rs
reuvenpo: done

cosmwasm/packages/wasmi-runtime/src/crypto/kdf.rs
reuvenpo: done

cosmwasm/packages/wasmi-runtime/src/crypto/aes_siv.rs
reuvenpo: done.
Oddly, rustfmt insisted on putting it above the comment. I have no idea why O.o

cosmwasm/packages/enclave-ffi-types/src/types.rs
Cashmaney: Lets make it ::default(). Just so it's consistent and slightly more clear what it's for

cosmwasm/packages/wasmi-runtime/src/wasm/db.rs
Cashmaney: Ah, sababa!

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
Cashmaney: Probably not worth the effort right now, just keep things like this in mind next time

Cashmaney:
> > Can you explain a little about what VmError and what UntrustedVmError are?
> 
> `VmError` is the standard error type for the `cosmwasm-sgx-vm` layer
> 
> > When do we need them?
> 
> During an ocall, we call into the original implementation of `db_read`, `db_write`, and `db_remove`. These call out all the way to the Go side. They return `VmError` when something goes wrong in this process.
> 
> > What kind of errors are they passing, and why they are important?
> 
> This mostly involves errors around FFI (memory errors, format errors, panic, etc.) but most importantly it communicates cases where an out-of-gas event happens on the Go side, which needs to trigger finalization of the computation and unwind everything all the way back to the enclave, out of the enclave, and back to Go. Some of the type conversions in this PR forward this signal across the enclave boundaries. On the `cosmwasm-vm` side this is already implemented.
> 
> > Also, why did you choose to make them wrapper pointers instead of just an error code (ala sgx_status_t)?
> 
> `VmError` is not FFI safe, and isn't designed to be. Instead of reimplementing it, which would affect a ton of code, I just box it before returning from the ocall, and unbox it when returning from the ecall.
> 
> Specifically for the out-of-gas situation, it has two sources. One is in the cosmos SDK. In our codebase this is the instance of `VmError` which gets boxed and unboxed. The second is when we see that the sum of gas consumed inside and outside the enclave exceeds the allowed limit for the computation. in that case we signal a `WasmEngineError::OutOfGas` [which is converted to](https://github.com/enigmampc/SecretNetwork/blob/8806932eebb53e03dae995c08c0a6a398ce0f554/cosmwasm/packages/wasmi-runtime/src/wasm/errors.rs#L37) `EnclaveError::OutOfGas` [which is converted to](https://github.com/enigmampc/SecretNetwork/blob/8806932eebb53e03dae995c08c0a6a398ce0f554/cosmwasm/packages/sgx-vm/src/errors.rs#L348) `VmError::GasDepletion`

Cool. Thanks for the explanation. Copy-paste this into the code somewhere so it doesn't go to waste.

Follow-up question - is there any risk to passing around a raw pointer into and out of the enclave vs. converting VmError to lets say.. an error code?

reuvenpo:
>Copy-paste this into the code somewhere so it doesn't go to waste

I'll add a part of this to the module containing `EnclaveError`

>is there any risk to passing around a raw pointer into and out of the enclave vs. converting VmError to an error code?

No. we never dereference the pointer inside the enclave (and we couldn't if we wanted to, it's pointing outside its memory space), so whatever the host gives us we will just forward back to the host. This is equivalent to what we do with the `context: Ctx` parameter.

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
reuvenpo: alright :+1:

cosmwasm/packages/wasmi-runtime/src/wasm/errors.rs
```diff
@@ -1,45 +1,27 @@
...
     EncryptionError,
     DecryptionError,
-    DbError(DbError),
+
     MemoryAllocationError,
     MemoryReadError,
     MemoryWriteError,
-    InputInvalid,
+
     InputEmpty,
```
reuvenpo, line 20: Looking at how this error variant is used, i think it should also return an error into the contract.

cosmwasm/packages/wasmi-runtime/src/wasm/errors.rs
```diff
@@ -1,45 +1,27 @@
...
     OutOfGas,
+    Panic,
+
     EncryptionError,
     DecryptionError,
-    DbError(DbError),
+
     MemoryAllocationError,
     MemoryReadError,
     MemoryWriteError,
```
reuvenpo, line 18: These three variants are really nice and high-level, but they eventually just get ignored. (converted to `EnclaveError::Unknown`). I'd like to instead use variants that are analogous to the variants of `CommunicationError` (see under the `cosmwasm-vm` errors), add similar variants to `EnclaveError`, and then add the two conversions. This way we'll have better integration with the error variants that are already available, and give better diagnostics for misbehaving contracts. What do you think? this is about 20 minutes of work, that i'd gladly do, but i want feedback on it.

Cashmaney: Lets open a new issue and do it some other time - we need to get this merged already

reuvenpo: https://github.com/enigmampc/SecretNetwork/issues/331

cosmwasm/packages/wasmi-runtime/src/wasm/errors.rs
reuvenpo: https://github.com/enigmampc/SecretNetwork/issues/329

PR #322, opened 2020-06-28
implement stub for query_chain

### LINKED ISSUE #321: Implement stub for query_chain

assafmo:
Closes #321

### DIFF
--- a/x/compute/internal/keeper/keeper_test.go
+++ b/x/compute/internal/keeper/keeper_test.go
@@ -200,8 +200,6 @@ func TestCreateWithGzippedPayload(t *testing.T) {
 }
 
 func TestInstantiate(t *testing.T) {
-	t.SkipNow() // escrow uses external query
-
 	tempDir, err := ioutil.TempDir("", "wasm")
 	require.NoError(t, err)
 	defer os.RemoveAll(tempDir)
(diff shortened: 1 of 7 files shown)


x/compute/internal/keeper/keeper_test.go
```diff
@@ -305,7 +303,7 @@ func TestExecute(t *testing.T) {
 
 	addr, err := keeper.Instantiate(ctx, contractID, creator, nil, initMsgBz, "demo contract 3", deposit)
 	require.NoError(t, err)
-	require.Equal(t, "cosmos18vd8fpwxzck93qlwghaj6arh4p7c5n89uzcee5", addr.String())
+	require.Equal(t, "enigma18vd8fpwxzck93qlwghaj6arh4p7c5n89d2p9uk", addr.String())
```
reuvenpo, line 306: isn't this supposed to start with `secret1`?

x/compute/internal/keeper/keeper_test.go
```diff
@@ -324,18 +322,20 @@ func TestExecute(t *testing.T) {
...
 
 	// verifier can execute, and get proper gas amount
 	start := time.Now()
 	gasBefore := ctx.GasMeter().GasConsumed()
 
-	res, err = keeper.Execute(ctx, addr, fred, msgBz, topUp)
+	msgBz, err := wasmCtx.Encrypt([]byte(`{"approve":{}}`))
+	require.NoError(t, err)
+
+	res, err := keeper.Execute(ctx, addr, fred, msgBz, topUp)
```
reuvenpo, line 338: I'm guessing that this test doesn't trigger the error you defined in this PR? At first I expected to see it checked here.

x/compute/internal/keeper/keeper_test.go
assafmo: Not just yet... (#320)

x/compute/internal/keeper/keeper_test.go
assafmo: Correct. It was crashing inside `keeper.Instantiate` with `Error in instantiation: Function("host module doesn\'t export function with name query_chain")`

assafmo:
> Looks good :)
> wouldn't we prefer to return a null buffer to the contract instead of crashing it though?

That's what I did at first, but then I thought it's better to crash it than to silently fail.

PR #345, opened 2020-07-01
TestAllocateOnHeap + enforce heap limit of 12 MiB per WASM VM

assafmo:
We are now enforcing 12 MiB (192 pages of 64 KiB) max heap size for WASM.

Tests:
1. Allocate 13 MiB at runtime, **the allocator fails**. 200,000 gas units. Enclave heap is 200 MiB.
2. Allocate 1 GiB at runtime, **get out of gas**. 200,000 gas units. Enclave heap is 200 MiB. This fails with `out of gas` because 1 GiB will ask for 134,217,728 gas units (8192 per page times 16,384 pages). The gas of 200,000 translates into 20,000,000 WASM gas units, so before the `memory_grow` opcode is reached the gas metering sees a request that will cost 134mn while the limit is 20mn, so it throws an `out of gas` exception.
3. Allocate 1 GiB at runtime, **the allocator fails**. 9,000,000 gas units (the max in CosmWasm). Enclave heap is 200 MiB. This time the WASM gas limit is 900mn which is more than 134mn from the previous test. We won't get `out of gas` but the allocator should fail because ! GiB is more than 12 MiB. This case is meant to test what happens if we try to allocate more than SGX has (200 MiB right now), but without running out of gas.


assafmo:
https://riot.im/app/#/room/!OoAWoTkXgxMLnLwGPO:matrix.parity.io/$15936050008132OEVfD:matrix.org

assafmo:
https://github.com/apache/incubator-teaclave-sgx-sdk/issues/254

reuvenpo:
> https://riot.im/app/#/room/!OoAWoTkXgxMLnLwGPO:matrix.parity.io/$15936050008132OEVfD:matrix.org

This requires a riot.im login. Can you please quote the comment you were pointing at?

cosmwasm/packages/wasmi-runtime/src/wasm/contract_operations.rs
```diff
@@ -214,7 +214,27 @@ fn start_engine(
...
+    let memory_entry = memory_section
+        .entries()
+        .first()
+        .ok_or(EnclaveError::CannotInitializeWasmMemory)?;
+
+    let minimum = std::cmp::max(17, memory_entry.limits().initial()); // no more than 17 (rust compiles it with 17, why???)
+    *memory_section.entries_mut() = vec![parity_wasm::elements::MemoryType::new(
+        minimum,
+        Some(192 /* 12 MiB */),
+    )];
```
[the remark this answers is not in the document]
reuvenpo: oh, you already pushed that change, cool :)
3f94a4a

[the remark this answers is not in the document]
assafmo: Done

x/compute/internal/keeper/testdata/test-contract/Cargo.toml
```diff
@@ -33,13 +33,8 @@ backtraces = ["cosmwasm-std/backtraces"]
 [dependencies]
 cosmwasm-std = { path = "../../../../../../cosmwasm/packages/std" }
 cosmwasm-storage = { path = "../../../../../../cosmwasm/packages/storage" }
-serde-json-wasm = { version = "0.2.1" }
 schemars = "0.7"
-serde = { version = "1.0.103", default-features = false, features = [
+serde = { version = "1.0.114", default-features = false, features = [
```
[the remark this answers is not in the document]
assafmo: It's in the test contract...

assafmo:
> > https://riot.im/app/#/room/!OoAWoTkXgxMLnLwGPO:matrix.parity.io/$15936050008132OEVfD:matrix.org
> 
> This requires a riot.im login. Can you please quote the comment you were pointing at?

@reuvenpo 

image

cosmwasm/packages/wasmi-runtime/src/wasm/contract_operations.rs
```diff
@@ -214,7 +215,25 @@ fn start_engine(
...
+        return Err(EnclaveError::CannotInitializeWasmMemory);
+    }
+
+    let memory_entry = memory_section
+        .entries()
+        .first()
+        .ok_or(EnclaveError::CannotInitializeWasmMemory)?;
+
+    let maximum: u32 = 192; // 12 MiB
+    let minimum: u32 = std::cmp::max(maximum, memory_entry.limits().initial());
```
assafmo, line 235: @reuvenpo Is this right?

reuvenpo: perfect :+1: 
I would even add a "contract requests too much initial memory" error if `memory_entry.limits().initial() > maximum`. you can also move `maximum` to a global `const`

assafmo: I think this should be `min` instead of `max` :thinking:

reuvenpo: you are correct, i did a stupid.

assafmo:
image

cosmwasm/packages/wasmi-runtime/src/wasm/contract_operations.rs
```diff
@@ -214,10 +215,59 @@ fn start_engine(
...
+        return Err(EnclaveError::CannotInitializeWasmMemory);
+    }
+
+    *memory_section.entries_mut() = vec![MemoryType::new(
+        requested_initial_pages,
+        Some(maximum_allowed_pages),
+    )];
+
+    trace!("Validated WASM memory demands");
+
```
Cashmaney, line 270: This seems like the type of thing that would be cleaner to separate to a different file/class

assafmo: It looks big because all the `map_err` and `error!()` out there, and I really don't like the amount of logic wrapping we are doing as it is, but your call...

Cashmaney: The point is to reduce the amount of information thrown at someone trying to understand what your code is doing. In this case you're doing memory initialization in the function "start_engine" which isn't trivial. You can move all the memory stuff into a different function at the very least, but for me personally it's not intuitive that memory initialization of the contract is in a file called "contract_operations.rs". Moving this stuff to "memory.rs" I think would make things clearer

assafmo: Done 5b6c07c

PR #351, opened 2020-07-02
Update ledger-nano-s.md

moonstash:
Added section for adding to light client so this page is more noob friendly.


assafmo:
Thanks @moonstash! Please see my previous comment. We want to avoid docs duplication as it gets harder to keep it updated over time.

moonstash:
@assafmo thanks for the feedback. I agree with you :)

assafmo:
Why did you close the PR? If you'd change it to link to the other doc we'll merge this.

PR #356, opened 2020-07-02
Report gas used on enclave error

reuvenpo:
This PR ensures that gas is reported from the enclave even if contract execution fails.

Parts of this PR are really ugly. I will comment on this as a code review.


cosmwasm/packages/wasmi-runtime/src/exports.rs
```diff
@@ -54,30 +54,33 @@ pub unsafe extern "C" fn ecall_init(
...
+        result_init_success_to_initresult(result, used_gas)
     });
     if let Ok(res) = result {
         res
     } else {
         error!("Call ecall_init panicked unexpectedly!");
         InitResult::Failure {
             err: EnclaveError::Panic,
+            // here we have no choice but to return 0, since the enclave panicked :(
+            used_gas: 0,
```
reuvenpo, line 83: Ok, this is a lie. we could also return the gas limit but then the user will likely observe an "out of gas" which isn't necessarily true, or useful.

What do you think?

cosmwasm/packages/wasmi-runtime/src/wasm/contract_operations.rs
```diff
@@ -67,34 +68,44 @@ pub fn init(
...
+    })?;
 
+    *used_gas = engine.gas_used();
     // todo: can move the key to somewhere in the output message if we want
     Ok(InitSuccess {
         output,
-        used_gas: engine.gas_used(),
         signature: contract_key,
     })
 }
```
reuvenpo, line 103: Ok so I hate this. I did it to not repeat this code:
```rust
.map_err(|err| {
    *used_gas = engine.gas_used();
    err
})
```
which has to be performed on each fallible operation from the call into the module until the return from this function.
We have two alternatives to this:

### Local variable and manually `map_err` for each fallible expression

In which case the code will look like this:
```rust
let report_gas = |err| {
    *used_gas = engine.gas_used();
    err
};

let vec_ptr = engine
    .init(env_ptr, msg_ptr)
    .map_err(wasmi_error_to_enclave_error)
    .map_err(report_gas)?;

let output = engine
    .extract_vector(vec_ptr)
    .map_err(wasmi_error_to_enclave_error)
    .map_err(report_gas)?;

trace!("Init output before encryption: {:?}", output);

// TODO: copy cosmwasm's structures to enclave
// TODO: ref: https://github.com/CosmWasm/cosmwasm/blob/b971c037a773bf6a5f5d08a88485113d9b9e8e7b/packages/std/src/init_handle.rs#L129
// TODO: ref: https://github.com/CosmWasm/cosmwasm/blob/b971c037a773bf6a5f5d08a88485113d9b9e8e7b/packages/std/src/query.rs#L13
let output =
    encrypt_output(output, secret_msg.nonce, secret_msg.user_public_key).map_err(report_gas)?;

trace!("Init output after encryption: {:?}", output);

*used_gas = engine.gas_used();
// todo: can move the key to somewhere in the output message if we want
Ok(InitSuccess {
    output,
    signature: contract_key,
})
```

### wrap the pattern into a macro

In which case the code will look like this:
```rust
// This at the top of the file or even in `lib.rs`
macro_rules! coalesce_err {
    ($error_type: ty, $body: expr) => {{
        #[allow(unused_mut)]
        let mut wrapper = || -> std::result::Result<_, $error_type> { Ok($body) };
        wrapper()
    }};
}

// and in the function:

// This wrapper is used to coalesce all errors in this block to one object
// so we can `.map_err()` in one place for all of them
let output = coalesce_err!(EnclaveError, {
    let vec_ptr = engine
        .init(env_ptr, msg_ptr)
        .map_err(wasmi_error_to_enclave_error)?;

    let output = engine
        .extract_vector(vec_ptr)
        .map_err(wasmi_error_to_enclave_error)?;

    trace!("Init output before encryption: {:?}", output);

    // TODO: copy cosmwasm's structures to enclave
    // TODO: ref: https://github.com/CosmWasm/cosmwasm/blob/b971c037a773bf6a5f5d08a88485113d9b9e8e7b/packages/std/src/init_handle.rs#L129
    // TODO: ref: https://github.com/CosmWasm/cosmwasm/blob/b971c037a773bf6a5f5d08a88485113d9b9e8e7b/packages/std/src/query.rs#L13
    let output = encrypt_output(output, secret_msg.nonce, secret_msg.user_public_key)?;

    trace!("Init output after encryption: {:?}", output);
    output
})
.map_err(|err| {
    *used_gas = engine.gas_used();
    err
})?;

*used_gas = engine.gas_used();
// todo: can move the key to somewhere in the output message if we want
Ok(InitSuccess {
    output,
    signature: contract_key,
})
```

The technique used in the PR, and the second alternative both allow marking a region in which the gas will always be reported. We can even move the free `*used_gas = engine.gas_used();` line inside the marked block. The macro option is even nicer in my opinion because it does `Ok`-wrapping, but in both cases a `return` applied to the block, not the whole function.
The first alternative is a lot simpler, but error prone if someone adds a `?` expression without `.map_err(report_gas)`.

cosmwasm/packages/wasmi-runtime/src/exports.rs
assafmo: I vote to return the limit.

cosmwasm/packages/wasmi-runtime/src/exports.rs
```diff
@@ -138,25 +144,28 @@ pub unsafe extern "C" fn ecall_query(
...
+        result_query_success_to_queryresult(result, used_gas)
     });
     if let Ok(res) = result {
         res
     } else {
         error!("Call ecall_query panic'd unexpectedly!");
         QueryResult::Failure {
             err: EnclaveError::Panic,
+            // here we have no choice but to return 0, since the enclave panicked :(
+            used_gas: 0,
```
assafmo, line 168: Also, return the limit.

cosmwasm/packages/wasmi-runtime/src/wasm/contract_operations.rs
assafmo: We used `map_err` + `?` everywhere else in the code, but I don't really have a strong opinion on this.
I think in the first case the compiler won't let you use `let` without `?`.

assafmo:
Just pointing out the only fail test:
```
2020-07-02T21:09:08.7839003Z 2020-07-02 21:09:08,783 TRACE [cosmwasm_sgx_vm::wasmi::wrapper] handle() called with env: "{\"block\":{\"height\":1234567,\"time\":1587556800,\"chain_id\":\"\"},\"message\":{\"sender\":\"xuM0y5WQirHcYTj9Dn02IP++goM=\",\"sent_funds\":[]},\"contract\":{\"address\":\"Oxp0hcYWLFiD7kX7LXR3qH2KTOU=\"},\"contract_key\":\"wknnEVH/+UkHXZcbz9PDDJZ+TP5wH70RLtwq5y14gCzseJTap97yYo8Y/cnVpWxM8bzQheA8NNiEbGrblTqkMw==\"}" msg: "0D�\u{e}ze\u{8}W2�v�,�\u{1f}��bN��<\u{13}\u{10}\u{2}��U��v���O�\u{3}n�,0\u{1a}��\u{17}�\u{8}��\t���\u{15}�j��S�\u{1}�8�?�p��\u{f}\u{f}\u{7}��Q�\u{15}�#3HN��>D�����~Y�� \u{2}�%#\u{7f}=��\u{1}�:���Ug�{#Y�e" enclave_id: 214649580552194 gas_limit: 19623800
2020-07-02T21:09:09.6427631Z 2020-07-02 21:09:09,642 TRACE [cosmwasm_sgx_vm::wasmi::wrapper] handle() returned with gas_used: 134280251 (gas_limit: 19623800)
2020-07-02T21:09:09.6435507Z thread '<unnamed>' panicked at 'attempt to subtract with overflow', /home/runner/work/SecretNetwork/SecretNetwork/cosmwasm/packages/sgx-vm/src/wasmi/wrapper.rs:195:9
2020-07-02T21:09:09.6435843Z     TestAllocateOnHeapFailBecauseGasLimit: secret_contracts_test.go:1004: 
2020-07-02T21:09:09.6435993Z         	Error Trace:	secret_contracts_test.go:1004
2020-07-02T21:09:09.6436324Z         	Error:      	Not equal: 
2020-07-02T21:09:09.6436467Z         	            	expected: "execute wasm contract failed: Ran out of gas"
2020-07-02T21:09:09.6436618Z         	            	actual  : "execute wasm contract failed: Caught Panic"
2020-07-02T21:09:09.6436755Z         	            	
2020-07-02T21:09:09.6437316Z         	            	Diff:
2020-07-02T21:09:09.6437730Z         	            	--- Expected
2020-07-02T21:09:09.6438042Z         	            	+++ Actual
2020-07-02T21:09:09.6438356Z         	            	@@ -1 +1 @@
2020-07-02T21:09:09.6438839Z         	            	-execute wasm contract failed: Ran out of gas
2020-07-02T21:09:09.6438999Z         	            	+execute wasm contract failed: Caught Panic
2020-07-02T21:09:09.6439142Z         	Test:       	TestAllocateOnHeapFailBecauseGasLimit
2020-07-02T21:09:09.6439634Z --- FAIL: TestAllocateOnHeapFailBecauseGasLimit (1.81s)
```
I would expect this to remain "out of gas".

Also, now that I think about it, I need to go back and fix every test that results in a failure and check that gas was indeed consumed by the enclave.

assafmo:
> Also, now that I think about it, I need to go back and fix every test that results in a failure and check that gas was indeed consumed by the enclave.

I actually can add that check only in the `{query,init,exec}Helper` functions and it'll cover all tests.

cosmwasm/packages/sgx-vm/src/wasmi/wrapper.rs
```diff
@@ -133,18 +133,20 @@ where
...
+            InitResult::Failure { used_gas, .. } => used_gas,
+        };
+
+        trace!(
+            target: module_path!(),
+            "init() returned with gas_used: {} (gas_limit: {})",
+            used_gas,
+            self.gas_left
+        );
+        self.gas_left -= used_gas;
```
reuvenpo, line 147: @assafmo I think the panic comes from here because now when we run out of gas it tries to subtract the gas_left below 0, but it's a unsigned integer and that panics in debug mode. (that's a good thing). I'll refactor this to either use `saturating_sub` or replace `gas_left` with `gas_used` so we can go higher. I'd still use `saturating add` but we won't lose information about how much gas was used unless someone actually uses more than `u64::MAX` gas points.
This will be very similar to what we do in `ContractInstance::use_gas` inside the enclave.

reuvenpo: yeah it even says in the logs:
```
 thread '<unnamed>' panicked at 'attempt to subtract with overflow', /home/runner/work/SecretNetwork/SecretNetwork/cosmwasm/packages/sgx-vm/src/wasmi/wrapper.rs:195:9
```

reuvenpo: yup, tests pass for 08d837e

cosmwasm/packages/wasmi-runtime/src/exports.rs
Cashmaney: The limit is set by the caller, no? So returning the limit could charge a whole lot of gas of someone manually set the limit, not knowing an error could cause the entire thing to be charged. If someone was sophisticated they would set a super low gas limit anyway, so this probably wouldn't be a big deterrent. I'd rather keep it at 0 for now, and think about this more as we go

cosmwasm/packages/wasmi-runtime/src/wasm/contract_operations.rs
Cashmaney: I'm fine with the macro, as long it's properly documented for those of us who aren't super familiar with macro syntax. Also, as long as we don't think that we will need to handle multiple error types (i.e. in the future we may want to return not only `EnclaveError`) and that it will be reused in other parts of the code.

Question - if we're adding the gas used every time we error out, why not return it as a part of the function call itself, rather than adding it after the fact? I mean change handle()/init/etc. to return Result<u32, (InterpreterError, GasUsed)> [or even create a new error type that contains both the error and the gas]
Another (unrelated to this PR) question: if we're mapping the error from handle()/etc/ to wasmi error anyway why aren't we returning EnclaveError in the first place?

cosmwasm/packages/wasmi-runtime/src/exports.rs
reuvenpo: seems like we have consensus on the 0. I'll keep it like this for now.

cosmwasm/packages/wasmi-runtime/src/wasm/contract_operations.rs
reuvenpo: As discussed - I'll switch this out to use the macro. Cleaning up the error handling will come soon after but as a separate PR.

PR #358, opened 2020-07-05
System tests in docker - local + azure

toml01:
* Refactored docker files a bit 
* Softened quote verification for non-production builds
* Some Makefile fixes around `FEATURES`


Makefile
```diff
@@ -197,12 +197,19 @@ build-testnet:
...
+	docker build --build-arg FEATURES=${FEATURES} --build-arg SGX_MODE=${SGX_MODE} -f Dockerfile.base -t rust-go-base-image .
+	docker build --build-arg --build-arg SGX_MODE=${SGX_MODE} --build-arg SECRET_NODE_TYPE=BOOTSTRAP -t enigmampc/secret-network-bootstrap-${ext}:${DOCKER_TAG} .
 
 docker_node:
-	docker build --build-arg SECRET_NODE_TYPE=NODE -t enigmampc/secret_node .
+	docker build --build-arg FEATURES=${FEATURES} --build-arg SGX_MODE=${SGX_MODE} -f Dockerfile.base -t rust-go-base-image .
+	docker build --build-arg --build-arg SGX_MODE=${SGX_MODE} --build-arg SECRET_NODE_TYPE=NODE -t enigmampc/secret-network-node-${ext}:${DOCKER_TAG} .
+
+docker_local_azure_hw:
+	docker pull cashmaney/aesm
```
Cashmaney, line 208: Don't think you need this one. It should pull automatically

Makefile
```diff
@@ -197,12 +197,19 @@ build-testnet:
 	docker build --build-arg SGX_MODE=HW --build-arg SECRET_NODE_TYPE=NODE -f Dockerfile.testnet -t enigmampc/secret-network-node:testnet .
 
 docker_bootstrap:
-	docker build --build-arg SGX_MODE=${SGX_MODE} --build-arg SECRET_NODE_TYPE=BOOTSTRAP -t enigmampc/secret-network-bootstrap-${ext}:${DOCKER_TAG} .
+	docker build --build-arg FEATURES=${FEATURES} --build-arg SGX_MODE=${SGX_MODE} -f Dockerfile.base -t rust-go-base-image .
```
Cashmaney, line 200: Lets separate this into it's own command so we don't have to duplicate this and line 204

cosmwasm/packages/wasmi-runtime/src/registration/cert.rs
```diff
@@ -347,6 +341,40 @@ pub fn verify_ra_cert(cert_der: &[u8]) -> SgxResult<Vec<u8>> {
...
+            warn!("TCB level of SGX platform service is outdated. You should check for firmware updates");
+            Ok(())
+        }
+        _ => {
+            error!("Invalid attestation quote status - cannot verify remote node");
+            Err(sgx_status_t::SGX_ERROR_UNEXPECTED)
+        }
+    }
+}
+
```
Cashmaney, line 377: :ok_hand:

cosmwasm/packages/wasmi-runtime/src/registration/cert.rs
```diff
@@ -347,6 +341,40 @@ pub fn verify_ra_cert(cert_der: &[u8]) -> SgxResult<Vec<u8>> {
...
 }
 
+#[cfg(all(feature = "SGX_MODE_HW", feature = "production"))]
+fn verify_quote_status(quote_status: SgxQuoteStatus) -> Result<(), sgx_status_t> {
+    match quote_status {
+        SgxQuoteStatus::OK => Ok(()),
+        SgxQuoteStatus::SwHardeningNeeded => {
+            warn!("Attesting enclave is vulnerable, and should be patched");
+            Ok(())
+        }
```
reuvenpo, line 351: wait, we DON'T want to crash on this in production??

cosmwasm/packages/wasmi-runtime/src/registration/cert.rs
```diff
@@ -347,6 +341,40 @@ pub fn verify_ra_cert(cert_der: &[u8]) -> SgxResult<Vec<u8>> {
...
+        _ => {
+            error!("Invalid attestation quote status - cannot verify remote node");
```
reuvenpo, line 353: I would add the reason to this log. doing `{:?}` on the `_` (after renaming it) should work.

cosmwasm/packages/wasmi-runtime/src/registration/cert.rs
```diff
@@ -347,6 +341,40 @@ pub fn verify_ra_cert(cert_der: &[u8]) -> SgxResult<Vec<u8>> {
     Ok(report_public_key)
 }
 
+#[cfg(all(feature = "SGX_MODE_HW", feature = "production"))]
```
reuvenpo, line 344: isn't this feature supposed to be added to `Cargo.toml`?

cosmwasm/packages/wasmi-runtime/src/registration/cert.rs
toml01: @Cashmaney you got blamed 😛 
Should this really give a warning  on production or should we crash instead?

toml01: Intel's documentation on this status, for reference:
```
"SWHardeningNeeded" - TCB level of the SGX platform is up-to-date but due to certain issues affecting the platform, additional SW Hardening in the attesting SGX enclaves may be needed.
```
https://api.portal.trustedservices.intel.com/documentation

or -
```
SW_HARDENING_NEEDED – the EPID signature of the ISV enclave QUOTE has been verified correctly but due to certain issues affecting the platform, additional SW Hardening in the attesting SGX enclaves may be needed. The relying party should evaluate the potential risk of an attack leveraging the relevant issues on the attesting enclave, and whether the attesting enclave employs adequate software hardening to mitigate the risk.
```
https://api.trustedservices.intel.com/documents/sgx-attestation-api-spec.pdf

toml01: Based on this:
https://cacheoutattack.com/files/SGAxe.pdf - section *Attestation Protocol Versions*
> As Intel has no way of
> remotely verifying if such hardening was perform or not, on
> machines vulnerable to LVI (nearly all currently available Intel
> machines), version 4 of the attestation protocol outputs SW -
> HARDENING NEEDED which indicates that our malicious
> quoting enclave’s “has been verified correctly but due to certain issues affecting the platform, additional SW Hardening in
> the attesting SGX enclaves may be needed” [48, Page 22]. Intel
> further instructs “the relying party should evaluate...whether
> the attesting enclave employs adequate software hardening
> to mitigate the risk”. This indicates that Intel asks enclave
> providers to trust their own enclaves only if they have performed LVI mitigations [42], without being able to remotely
> verify so. 

I'd say we don't want that in production.

cosmwasm/packages/wasmi-runtime/src/registration/cert.rs
toml01: Do you refer to the `production` feature? It's already in:
https://github.com/enigmampc/SecretNetwork/blob/system-tests-in-docker/cosmwasm/packages/wasmi-runtime/Cargo.toml#L16

cosmwasm/packages/wasmi-runtime/src/registration/cert.rs
Cashmaney: Right now there's no microcode fix for the LVI issue, and so it will always return SW_HARDENING_NEEDED

toml01: Ahh, so SW in this context always refers to our code then?
Sounds legit. Maybe we should print the Advisory IDs as well in the warning.

cosmwasm/packages/wasmi-runtime/src/registration/cert.rs
toml01: :+1:

cosmwasm/packages/wasmi-runtime/src/registration/cert.rs
reuvenpo: oh, i must've missed it

PR #364, opened 2020-07-06
fix error decrypting on smart query

assafmo:
Fixes https://github.com/enigmampc/SecretNetwork/issues/365

- [x] Decrypt error in the response of a smart query
- [x] Export error decryption to the encryption utils
- [x] Also print data as string in `secretcli compute q tx`


x/compute/client/cli/query.go
```diff
@@ -444,13 +444,30 @@ func GetCmdGetContractStateSmart(cdc *codec.Codec) *cobra.Command {
 			if err != nil {
 				return err
 			}
+			nonce := queryData[:32]
 
 			res, _, err := cliCtx.QueryWithData(route, queryData)
 			if err != nil {
+				if strings.Contains(err.Error(), "wasm contract failed: generic: ") {
```
toml01, line 451: Wouldn't it be better to wrap it in a function?

assafmo: It would... 🤣

PR #367, opened 2020-07-07
Fix secretjs storage

assafmo:
SecretJS now requires tx sender seed in the constructor, instead of managing local storage.


levackt:
If you add seed as the last parameter to the constructor it should still work with other usages.

cosmwasm-js/packages/sdk/types/signingcosmwasmclient.d.ts
```diff
@@ -66,6 +66,7 @@ export declare class SigningCosmWasmClient extends CosmWasmClient {
     apiUrl: string,
     senderAddress: string,
     signCallback: SigningCallback,
+    seed?: Uint8Array,
```
levackt, line 69: I meant the seed could move to last over here, and be consistent with the rest client change above.

[And here's the ugly change I did to make it work for my sample](https://github.com/levackt/secret-counter/commit/dfba43ddd5028a65ae97dcc5c2fda209a141d6ad#diff-ecc0527dbc297f2f43790c6dd291e4ebR70), maybe I missed a simple TS thing as a noob, but still would be nice to be consistent if possible.

assafmo: Thanks, you're right.

PR #371, opened 2020-07-08
Update join-validator-mainnet.md

reuvenpo:
Slight change to a misleading heading


assafmo:
I think it's better to edit docs directly on master.

PR #387, opened 2020-07-12
Simplifying query commands

toml01:
closes #378

PR #392, opened 2020-07-13
Implement query_chain()

### LINKED ISSUE #277: Implement query_chain
External query allows contracts to call other contracts mid-execution. Not a priority for testnet

- [x] Implement naive `query_chain()`
- [x] Implement stack limit for queries

assafmo:
Closes #277.

- [x] External query
- [x] Graceful handling of OOM inside the enclave 
- [x] Fix a bug in x/compute that prevented a deeper call stack than 1 external query (see https://github.com/CosmWasm/wasmd/pull/213)
- [x] Fix gas isn't being consumed if the enclave panics (see https://github.com/CosmWasm/wasmd/pull/216)
- [x] Enable the `staking` feature


cosmwasm/packages/wasmi-runtime/src/cosmwasm/coins.rs
```diff
@@ -0,0 +1,9 @@
+use serde::{Deserialize, Serialize};
+
+use crate::cosmwasm::math::Uint128;
+
+#[derive(Serialize, Deserialize, Clone, Default, Debug, PartialEq)]
+pub struct Coin {
+    pub denom: String,
+    pub amount: Uint128,
+}
```
reuvenpo, line 9: For this file, and other files that we copied and modified from `cosmwasm-std`, please add a comment at the top of the file (with `//! lorem ipsum` syntax) mentioning the filename of the source, and a warning/request to keep the definitions in sync.

cosmwasm/packages/wasmi-runtime/src/wasm/errors.rs
```diff
@@ -60,7 +62,8 @@ impl From<WasmEngineError> for EnclaveError {
             MemoryAllocationError => EnclaveError::MemoryAllocationError,
             MemoryReadError => EnclaveError::MemoryReadError,
             MemoryWriteError => EnclaveError::MemoryWriteError,
-            NotImplemented => EnclaveError::NotImplemented,
+            BadQueryChainRequest => EnclaveError::BadQueryChainRequest,
+            BadQueryChainResponse => EnclaveError::BadQueryChainResponse,
             // Unexpected WasmEngineError variant
```
reuvenpo, line 67: So, in the cosmwasm implementation they return an error into the contract when this happens. This is especially important for `BadQueryChainResponse` because if contract B misbehaves, we don't want to just crash contract A.
Let's talk about this and thing about the best way to implement this.

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,136 @@
...
+use log::*;
+use sgx_types::sgx_status_t;
+
+pub fn encrypt_and_query_chain(
+    query: &[u8],
+    context: &Ctx,
+    nonce: IoNonce,
+    user_public_key: Ed25519PublicKey,
+) -> Result<(Option<Vec<u8>>, u64), WasmEngineError> {
+    // TODO encrypt query
```
reuvenpo, line 19: You do encrypt it, a couple lines below

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,136 @@
...
+                );
+
+                WasmEngineError::EncryptionError
+            })?;
+            let msg = Binary(encrypted_msg.to_slice());
+            QueryRequest::Wasm(WasmQuery::Smart { contract_addr, msg })
+        }
+        QueryRequest::Wasm(_) => query_struct, // WasmQuery::Raw - pass it to x/compute to fail because the state is encrypted on-chain
+        _ => query_struct,
+    };
```
reuvenpo, line 50: Ok so i was thinking, we can probably change this to something like:
```rust
if let QueryRequest::Wasm(WasmQuery::Smart { contract_addr, msg }) = &mut query_struct {
  // create the secret message
  // encrypt it
  msg = Binary(encrypted_msg.to_slice()); // without `let`. just assign into the original object.
}
```

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    let encrypted_query = serde_json::to_vec(&query_struct).map_err(|err| {
+        // this should never happen
+        error!(
+            "encrypt_and_query_chain() cannot parse json from struct {:?}: {:?}",
```
reuvenpo, line 55: "cannot create/generate/build json from struct"

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
```diff
@@ -200,9 +213,6 @@ impl WasmiApi for ContractInstance {
         let state_key_name = self
             .extract_vector(state_key_ptr_ptr as u32)
             .map_err(|err| {
-                // Assaf: I don't think this can happen unless the caller overrides the CosmWasm implementation of
-                // deps.api.canonical_address, not uses the CosmWasm std at all, writes the contract in a language
-                // other than rust, etc.
                 error!("read_db() error while trying to read state_key_name from wasm memory");
```
reuvenpo, line 216: why did you decide to remove these comment?

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
```diff
@@ -469,8 +452,44 @@ impl WasmiApi for ContractInstance {
...
+            error!("query_chain() error while trying to read canonical address from wasm memory",);
+            err
+        })?;
+
+        trace!(
+            "query_chain() was called from WASM code with {:?}",
+            String::from_utf8_lossy(&query_buffer)
+        );
+
+        // TODO pass query_buffer to ocall
```
reuvenpo, line 466: you completed this todo

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
```diff
@@ -469,8 +452,44 @@ impl WasmiApi for ContractInstance {
...
+        let (result, gas_used) = encrypt_and_query_chain(
+            &query_buffer,
+            &self.context,
+            self.user_nonce,
+            self.user_public_key,
+        )?;
+        self.use_gas_externally(gas_used)?;
+
+        let result = match result {
+            None => return Ok(Some(RuntimeValue::I32(0))), // Is this supposed to be 0 or Err?
```
reuvenpo, line 478: I think we should be returning errors in this case, but i'll check.

cosmwasm/packages/sgx-vm/src/wasmi/exports.rs
```diff
@@ -76,6 +77,59 @@ pub extern "C" fn ocall_read_db(
...
+                    unsafe { *gas_used = gas_cost };
+                    ans.map(|val| {
+                        super::allocate_enclave_buffer(&val.0).map_err(|_| OcallReturn::Failure)
+                    })
+                    .unwrap_or_else(|_| Ok(EnclaveBuffer::default()))
+                }
+                Ok((Err(_), gas_cost)) => {
+                    unsafe { *gas_used = gas_cost };
+                    Err(OcallReturn::Failure)
+                }
```
reuvenpo, line 114: So, in this context, you shouldn't match on the `SystemResult` as you do here, but rather you need to serialize it and return it into the enclave. The `Ok((Err(_), gas_cost))` is not a failure that this layer handles, but it's a failure that the contract handles.

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
assafmo: Because they are not correct, and this behavior is already being tested in the CI.

cosmwasm/packages/sgx-vm/src/wasmi/exports.rs
assafmo: But who consumes the gas?

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
reuvenpo: ah, awesome :+1:

cosmwasm/packages/sgx-vm/src/wasmi/exports.rs
reuvenpo: I'm not sure i communicated this correctly. instead of these two cases, you should have:
```rust
Ok((system_result, gas_cost)) => {
  unsafe { *gas_used = gas_cost };
  let response = json_serde::to_vector_or_whatever(system_result).unwrap_or_else(|| EnclaveBuffer::default());
  super::allocate_enclave_buffer(&response).map_err(|_| OcallReturn::Failure)
}
```

assafmo: c7fc865cf446a4a0d657758ab2ba5f4a

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
assafmo: Not an `Option` anymore, so this dilemma no longer exist.

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    if !is_encrypted {
+        return answer;
+    }
+
+    // Call query_chain (this bubbles up to x/compute via ocalls and FFI to Go code)
+    // This returns the answer from x/compute
+    match answer {
+        Ok((encrypted_answer_as_vec, gas_used)) => {
+            // answer is QueryResult (Result<Binary,StdError>) encoded be serde to bytes
```
reuvenpo, line 78: This match can be replaced by:
```rust
let (encrypted_answer_as_vec, gas_used) = answer?;
// The main arm of the match here.
```
this will also help flatten the code (but not The Curve, unfortunately)

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
```diff
@@ -469,8 +452,47 @@ impl WasmiApi for ContractInstance {
...
+            self.user_public_key,
+        )?;
+        trace!(
+            "query_chain() got answer from outside with result {:?}",
+            String::from_utf8_lossy(&answer)
+        );
+        trace!(
+            "query_chain() got answer from outside with gas {}",
+            gas_used
+        );
```
reuvenpo, line 481: can this be reduced to one `trace!()`?

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
```diff
@@ -469,8 +452,47 @@ impl WasmiApi for ContractInstance {
...
+
+        self.use_gas_externally(gas_used)?;
+
+        let ptr_to_region_in_wasm_vm =   self.write_to_memory(&answer)
```
reuvenpo, line 485: extra spaces here.

_I swear I can solve captchas._

x/compute/internal/keeper/testdata/test-contract/src/contract.rs
```diff
@@ -239,9 +237,45 @@ pub fn handle<S: Storage, A: Api, Q: Querier>(
...
+    let err = deps
+        .querier
+        .query::<u8>(&QueryRequest::Wasm(WasmQuery::Smart {
+            contract_addr,
+            msg: Binary(r#"{"panic":{}}"#.as_bytes().to_vec()),
+        }))
+        .unwrap_err();
+
+    Err(err)
+}
```
reuvenpo, line 277: nice code reuse haha, calling yourself like that is cool

x/compute/module_test.go
```diff
@@ -189,11 +187,11 @@ func TestHandleInstantiate(t *testing.T) {
...
-	assertContractState(t, q, data.ctx, contractAddr, state{
-		Verifier:    []byte(fred),
-		Beneficiary: []byte(bob),
-		Funder:      []byte(creator),
-	})
+	// assertContractState(t, q, data.ctx, contractAddr, state{
+	// 	Verifier:    []byte(fred),
+	// 	Beneficiary: []byte(bob),
+	// 	Funder:      []byte(creator),
+	// })
```
reuvenpo, line 194: why did you comment stuff out in this file?

assafmo: It's a remnant of #387...

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
assafmo: It's the effing rustfmt and macros

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
assafmo: Yep

cosmwasm/packages/wasmi-runtime/src/wasm/runtime/contract.rs
reuvenpo: I figured :joy:

x/compute/module_test.go
reuvenpo: Oh I see, very well :+1:

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,232 @@
...
+        );
+        WasmEngineError::BadQueryChainRequest
+    })?;
+
+    let mut is_encrypted = false;
+
+    if let QueryRequest::Wasm(WasmQuery::Smart {
+        contract_addr: _,
+        msg,
+    }) = &mut query_struct
```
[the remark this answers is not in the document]
assafmo: tenor

cosmwasm/packages/wasmi-runtime/src/wasm/errors.rs
assafmo: It turns out we only use `BadQueryChainRequest`...

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,224 @@
...
+    })?;
+
+    let answer = query_chain(context, &encrypted_query);
+
+    if !is_encrypted {
+        return answer;
+    }
+
+    // Call query_chain (this bubbles up to x/compute via ocalls and FFI to Go code)
+    // This returns the answer from x/compute
```
Cashmaney, line 92: Isn't this a comment for `let answer = query_chain(context, &encrypted_query);`?

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,224 @@
...
+            }
+        },
+        Ok(Err(std_error)) => {
+            error!(
+                        "encrypt_and_query_chain() got an StdError as an answer, but it should be of type GenericErr and encrypted inside. Got instead: {:?}",
+                        std_error
+                    );
+            Ok(Err(std_error))
+        }
+    };
```
Cashmaney, line 189: Make a separate function(s). This entire thing is basically a `decrypt_answer` function, and inside you have smaller  functions, e.g. `decrypt_generic_err`. Would go a long way to making this easier to digest (you're returning stuff like ` Ok(Ok(Binary(decrypted)))` - this isn't easy to understand without the code being well structured)

cosmwasm/packages/wasmi-runtime/src/wasm/errors.rs
assafmo: Let me know what you think about 3f9b2da8a6ede4cfc32f8ea69a62dc7f316e49a0

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,224 @@
...
+use crate::cosmwasm::{
+    std_error::{StdError, StdResult},
+    system_error::SystemResult,
+};
+
+use enclave_ffi_types::{Ctx, EnclaveBuffer, OcallReturn, UntrustedVmError};
+use log::*;
+use sgx_types::sgx_status_t;
+
+pub fn encrypt_and_query_chain(
```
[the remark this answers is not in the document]
assafmo: Yeah yeah, I was just trying to make it pass the tests first

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    let encrypted_query = serde_json::to_vec(&query_struct).map_err(|err| {
+        // this should never happen
+        error!(
+            "encrypt_and_query_chain() cannot build json from struct {:?}: {:?}",
+            query_struct, err
+        );
+        WasmEngineError::BadQueryChainRequest
+    })?;
```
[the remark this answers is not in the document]
assafmo: It's a bit tough because there are 3 returns types from this code section. You're right though, Once all the tests pass I'll refactor this entire function. It's these damn `match` clauses :unamused:

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
assafmo: You're right. Will fix this once all tests pass.

The `Ok(Ok(...` stuff is part of CosmWasm's API for `Querier`.
The caller contract expects to receive `Result<Result<Binary, StdError>, SystemError>`.
- `Binary` - An encrypted answer from the callee.
- `StdError` - An encrypted error from the callee, or if panic a plaintext error.
- `SystemError` - A plaintext error from the VM of the callee. See [this](https://github.com/enigmampc/SecretNetwork/blob/7754e997655bb33d5d6f0614b280e4008a3630bd/cosmwasm/packages/wasmi-runtime/src/cosmwasm/system_error.rs#L8-L16) for more info.

cosmwasm/packages/sgx-vm/src/wasmi/exports.rs
```diff
@@ -76,6 +71,58 @@ pub extern "C" fn ocall_read_db(
...
+            match answer {
+                Ok((system_result, gas_cost)) => {
+                    unsafe { *gas_used = gas_cost };
+
+                    /*
+                        wasm code expects to get this as Result<Result<Binary, StdError>, SystemError> which is called SystemResult
+                        see CosmWasm's implementation https://github.com/enigmampc/SecretNetwork/blob/508e99c990dd656eb61f456584dab054487ba178/cosmwasm/packages/sgx-vm/src/imports.rs#L124
+                    */
+
+                    crate::serde::to_vec(&system_result)
```
reuvenpo, line 101: Can you please make this a regular `//` comment so it takes 2 lines instead of 5?

cosmwasm/packages/sgx-vm/src/wasmi/exports.rs
```diff
@@ -213,6 +262,24 @@ where
     })
 }
 
+// get
+// FfiResult<(Option<Vec<u8>>, u64)>
+// FfiResult<(SystemResult<StdResult<Binary>>, u64)>
+// query
+
+fn ocall_query_chain_impl<S, Q>(
```
reuvenpo, line 270: please remove this comment

cosmwasm/packages/enclave-ffi-types/src/types.rs
```diff
@@ -140,6 +140,9 @@ pub enum EnclaveError {
     /// Unexpected Error happened, no more details available
     #[display(fmt = "unknown error")]
     Unknown,
+
+    BadQueryChainRequest,
+    BadQueryChainResponse,
 }
```
reuvenpo, line 146: You should still remove these variants

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,278 @@
...
+                    "encrypt_and_query_chain() got an error while trying to serialize the error {:?} returned to WASM: {:?}",
+                    answer,
+                    err
+                );
+
+                WasmEngineError::SerializationError
+            })?;
+
+            return Ok((answer_as_vec, 500)); // Should we charge gas for this to prevent spam?
+        }
```
reuvenpo, line 48: >Should we charge gas for this to prevent spam?

I think it doesn't really matter since the contract will still have to run some code inside the vm between consecutive calls to imports. That being said, this can't hurt since users shouldn't encounter it in regular use.

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,278 @@
...
+    user_public_key: Ed25519PublicKey,
+) -> Result<(Vec<u8>, u64), WasmEngineError> {
+    let mut query_struct: QueryRequest = match serde_json::from_slice(query) {
+        Ok(query_struct) => query_struct,
+        Err(err) => {
+            error!(
+                "encrypt_and_query_chain() cannot build struct from json {:?}: {:?}",
+                String::from_utf8_lossy(query),
+                err
+            );
```
reuvenpo, line 30: This is `debug!()` at most. similarly below

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,278 @@
...
+                    }
+                }
+            }
+        },
+        Ok(Err(std_error)) => {
+            error!(
+                "encrypt_and_query_chain() got an StdError as an answer, but it should be of type GenericErr and encrypted inside. Got instead: {:?}",
+                std_error
+            );
+            Ok(Err(std_error))
```
reuvenpo, line 220: I'd say this is `warning!()`

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,278 @@
...
+
+    trace!(
+        "encrypt_and_query_chain() decrypted the answer to be: {:?}",
+        answer
+    );
+
+    let answer_as_vec = serde_json::to_vec(&answer).map_err(|err| {
+                error!("encrypt_and_query_chain() got an error while trying to serialize the decrypted answer to bytes: {:?}", err);
+                WasmEngineError::SerializationError
+            })?;
```
reuvenpo, line 232: indentation.

This `rustfmt` behaviour around macros is idiotic...

cosmwasm/packages/wasmi-runtime/src/wasm/errors.rs
reuvenpo: Ah yeah, I didn't comment - it's great :) :+1:

cosmwasm/packages/wasmi-runtime/src/exports.rs
```diff
@@ -214,11 +214,6 @@ pub unsafe extern "C" fn ecall_query(
     } else {
         *used_gas = gas_limit;
 
-        error!(
-            "ecall_query panic: used_gas {} gas_limit {}",
-            *used_gas, gas_limit
-        );
-
```
assafmo: Reuven found what caused this. `gas_used` in `Enclave.edl` is write-only, so trying to read it caused segfault
https://github.com/enigmampc/SecretNetwork/blob/6537eaead0cdd48bd2a06adc56107c20221107a9/cosmwasm/packages/wasmi-runtime/Enclave.edl#L74

cosmwasm/packages/wasmi-runtime/src/oom_handler.rs
```diff
@@ -0,0 +1,19 @@
+use core::sync::atomic::{AtomicBool, Ordering};
+
+// This is thread_local to prevent race conditions in case multiple
+// threads are running while OOM happens
+// Mutex won't work because you cannot hold the Mutex between throwing
+// and catching the panic
+static OOM_HAPPANED: AtomicBool = AtomicBool::new(false);
```
reuvenpo, line 7: This comment is outdated

cosmwasm/packages/wasmi-runtime/src/wasm/gas.rs
```diff
@@ -37,10 +37,10 @@ impl Default for WasmCosts {
             mem: 2,
             static_u256: 64,
             static_address: 40,
-            initial_mem: 4096,
+            initial_mem: 4096, // Assaf: I don't think this goes anywhere
             grow_mem: 8192,
             memcpy: 1,
-            max_stack_height: 64 * 1024,
+            max_stack_height: 64 * 1024, // Assaf: I don't think this goes anywhere
```
reuvenpo, line 43: what do you mean?

assafmo: By looking at how the gas metering works, this isn't used.

assafmo: Look at https://github.com/enigmampc/SecretNetwork/blob/1ff31db4940e5162c71a2691f98aac60a753336f/cosmwasm/packages/wasmi-runtime/src/wasm/gas.rs#L50-L72

reuvenpo: huh, that's weird...
I think we should investigate what this does - maybe we're using it wrong - but that's a separate issue.

assafmo: Oh I got it. I think `initial_mem` was being used as `regular`. Fixed it in 86d1ae1e8cc7b9e4080d66073fe95794b0e6814f

reuvenpo: ah, awesome! thanks for taking care of it!

assafmo: Asked parity about controlling the stack usage: https://matrix.to/#/!OoAWoTkXgxMLnLwGPO:matrix.parity.io/$1596011601124301ImAZX:matrix.org?via=matrix.parity.io&via=matrix.org

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,309 @@
...
+            "encrypt_and_query_chain() got an error while trying to serialize the error {:?} returned to WASM: {:?}",
+            answer,
+            err
+        );
+
+        WasmEngineError::SerializationError
+    })
+}
+
+fn system_error_invalid_response<T>(response: Vec<u8>, err: T) -> Result<Vec<u8>, WasmEngineError>
```
Cashmaney, line 204: Can you add a comment on what this error is and when it can happen?

cosmwasm/packages/sgx-vm/src/wasmi/wrapper.rs
```diff
@@ -103,9 +102,12 @@ where
...
     }
 
+    pub fn gas_used(&self) -> u64 {
+        self.used_gas
+    }
+
     pub fn init(&mut self, env: &[u8], msg: &[u8]) -> VmResult<InitSuccess> {
-        trace!(
-            target: module_path!(),
+        println!(
```
Cashmaney, line 110: I think all of these should be switched back to `debug` or `trace`?

cosmwasm/packages/wasmi-runtime/src/exports.rs
```diff
@@ -74,17 +94,26 @@ pub unsafe extern "C" fn ecall_init(
...
         result_init_success_to_initresult(result)
     });
     if let Ok(res) = result {
         res
     } else {
-        error!("Call ecall_init panicked unexpectedly!");
-        // The enclave panicked. we do not report gas used in this case. (it should be initialized to 0)
-        InitResult::Failure {
-            err: EnclaveError::Panic,
+        *used_gas = gas_limit / 2;
```
Cashmaney, line 105: Can you add a comment here why we are just doing / 2? Also, is this something we may want to tune? If so make it a constant in consts.rs.

Repeat for all the other copies of this line

cosmwasm/packages/wasmi-runtime/src/exports.rs
```diff
@@ -124,17 +155,27 @@ pub unsafe extern "C" fn ecall_handle(
     let env = std::slice::from_raw_parts(env, env_len);
     let msg = std::slice::from_raw_parts(msg, msg_len);
     let result = panic::catch_unwind(|| {
-        let used_gas_ref = &mut *used_gas;
-        let result = crate::wasm::handle(context, gas_limit, used_gas_ref, contract, env, msg);
+        let mut local_used_gas = *used_gas;
```
Cashmaney, line 158: Can you add a comment why you did it this way?

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -0,0 +1,309 @@
...
+        }
+        Ok(Err(std_error)) => {
+            // TODO when removing this you can replace this clause with: `other => other` instead of the `Ok(Err(...))`
+            debug!(
+                "encrypt_and_query_chain() got an StdError as an answer, but it should be of type GenericErr and encrypted inside. Got instead: {:?}",
+                std_error
+            );
+            Ok(Err(std_error))
+        }
+    };
```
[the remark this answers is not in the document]
assafmo: We tried. It's very hard because of the inner `return`s, and makes it much less clear to read.

cosmwasm/packages/wasmi-runtime/src/exports.rs
assafmo: Both works, but @reuvenpo and I thought that it improves readability.
It was hard figuring where these references came from while chasing gas issues.

cosmwasm/packages/wasmi-runtime/src/exports.rs
assafmo: Yeah. I wanted to consume a lot of gas for panics, but consuming all of it masks the real panic with an out of gas error.

Issue #377, opened 2020-07-11
Go over error messages and logs


Cashmaney:
Make sure they're accurate/make sense/don't have anything stupid/spam the log

- [x] CosmWasm error messages
- [x] Attestation logs
- [x] Enclave logs

PR #393, opened 2020-07-13
#377 go over errors

Cashmaney:
Main stuff:

+ Moved a lot of debug/info to trace so it doesn't clog up the stdout

+ Set the minimum log levels for the enclave as follows:

trace in SW mode
info in HW/non prod
warn in HW/prod

+ Changed error log text to be more human-readable. Examples:

wasm contract failed: generic: -> contract failed: encrypted:
FailedOcall -> failed to execute ocall

etc.


docs/post-mortems/testnet-halt-95.md
```diff
@@ -159,7 +159,7 @@
     ```
 8. Then a problem occured with the `compute` module:
     ```
-    panic: create wasm contract failed: Wasm Error: Filesystem error: File exists (os error 17)
+    panic: create contract failed: Wasm Error: Filesystem error: File exists (os error 17)
```
assafmo, line 162: Find and replace much? :rofl:

go-cosmwasm/types/stderror.go
```diff
@@ -62,7 +62,7 @@ type GenericErr struct {
 }
 
 func (e GenericErr) Error() string {
-	return fmt.Sprintf("generic: %s", e.Msg)
+	return fmt.Sprintf("encrypted: %s", e.Msg)
```
assafmo, line 65: I don't think this is right. The inner error will also be "encrypted" if you `.Error()` it.

x/compute/client/cli/query.go
```diff
@@ -369,20 +366,24 @@ func GetCmdQuery(cdc *codec.Codec) *cobra.Command {
...
 				}
+				// Itzik: Commenting this as it might have been a placeholder for encrypting
+				//else if strings.Contains(err.Error(), "EnclaveErr") {
+				//	return err
+				//}
 				return err
 			}
 
-			resDecrypted := []byte{}
+			var resDecrypted []byte
```
assafmo, line 386: Same same

go-cosmwasm/types/stderror.go
Cashmaney: Isn't this already the inner error? That's why I changed it in the first place

docs/post-mortems/testnet-halt-95.md
Cashmaney: 🤣

x/compute/client/cli/query.go
Cashmaney: Yep. It was just to silence an annoying IDE warning

go-cosmwasm/types/stderror.go
assafmo: In order to keep the go side happy, when we encrypt the output we wrap any error within a generic error.
This:
```json
{ "generic_err": { "msg": "bla bla" } }
```
Becomes this:
```json
{ "generic_err": { "msg": "{ \"generic_err\": { \"msg\": \"bla bla\" } }" } }
```
Becomes this:
```json
{ "generic_err": { "msg": "1GGPyFBJVaHRVnoc+0WP2Maf9ulTqCStJsVsDuj64PsyjaUEbldM94+z/dW730nOhFFaSnmmoEw=" } }
```
After decryption on the go side, we get the original error which is `generic_err`, but when we `.Error()` it we'll get this `encrypted:` string instead of the `generic:` string.

It's not that big of a deal...

assafmo: (Also lets pretend that the random bytes here weren't generated by `cat /dev/urandom | head -c 56 | base64`)

Cashmaney: Hmm... I changed it specfically because when outputting to the chain to change:

``` "raw_log": "execute wasm contract failed: generic: Jz0Nt/f/ZH9VWi1KQOL/WlWJ2ThcqFGFiEZ/1DOU2tEx40ubeWpGAHXtp3xb/ipBTGi/rQ==: failed to execute message; message index: 0",```

to:

``` "raw_log": "execute wasm contract failed: encrypted: Jz0Nt/f/ZH9VWi1KQOL/WlWJ2ThcqFGFiEZ/1DOU2tEx40ubeWpGAHXtp3xb/ipBTGi/rQ==: failed to execute message; message index: 0",```

Is there any other case where it would make things less clear? I tried searching the logs in the integration tests but didn't see another example

Comment (assafmo) [reply go-cosmwasm/types/stderror.go:65]: Probably CosmWasm are using this to show errors to users on the CLI.
Can you search where `StdError` is returned from the keeper and what the handler/CLI is doing with it?
It's most probably just returned as a string to the user, so I'd say I'm fine with it. :shrug:

Comment (Cashmaney) [reply go-cosmwasm/types/stderror.go:65]: Didn't find anything except returning as a string. Worst comes to worst we can change this later on as well

File: x/compute/internal/types/errors.go [hunk 4]
```diff
@@ -2,39 +2,59 @@ package types
...
+// ** Warning **
+// Below are functions that check for magic strings that depends on the output of the enclave.
+// Beware when changing this, or the rust error string
+func ContainsEnclaveError(str string) bool {
+	return strings.Contains(str, "Enclave")
+}
+
+func ContainsEncryptedString(str string) bool {
+	return strings.Contains(str, "encrypted: ")
+}
```
reuvenpo, line 60: I would change this to use something like "StartsWith" if possible. If the contract is able to inject plaintext information to the strings that get checked by these functions, then they can mess with this logic.

x/compute/internal/types/errors.go
```diff
@@ -2,39 +2,59 @@ package types
...
 	// ErrCreateFailed error for wasm code that has already been uploaded or failed
-	ErrCreateFailed = sdkErrors.Register(DefaultCodespace, 1, "create wasm contract failed")
-
-	// ErrAccountExists error for a contract account that already exists
-	ErrAccountExists = sdkErrors.Register(DefaultCodespace, 2, "contract account already exists")
+	ErrCreateFailed = sdkErrors.Register(DefaultCodespace, 1, "create contract failed")
 
 	// ErrInstantiateFailed error for rust instantiate contract failure
-	ErrInstantiateFailed = sdkErrors.Register(DefaultCodespace, 3, "instantiate wasm contract failed")
+	ErrInstantiateFailed = sdkErrors.Register(DefaultCodespace, 2, "instantiate contract failed")
```
reuvenpo, line 17: Why did you reshuffle these lines?

cosmwasm/packages/wasmi-runtime/src/registration/cert.rs
```diff
@@ -346,7 +346,7 @@ fn verify_quote_status(quote_status: SgxQuoteStatus) -> Result<(), sgx_status_t>
     match quote_status {
         SgxQuoteStatus::OK => Ok(()),
         SgxQuoteStatus::SwHardeningNeeded => {
-            warn!("Attesting enclave is vulnerable, and should be patched");
+            // warn!("Attesting enclave is vulnerable, and should be patched");
```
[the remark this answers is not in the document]
Cashmaney: Because it's not true for SW_HARDENING_NEEDED, and it prints it out not only locally, but for the entire chain every time someone registers

x/compute/internal/types/errors.go
Cashmaney: It's not startswith because these won't be at the start of the string. Also, all the inputs that users can create is encrypted afaik, so it's hard to make this act up. Finally, it only affects queries, so I would be more concerned about a mistake causing something weird to be displayed rather than malicious behaviour.

x/compute/internal/types/errors.go
Cashmaney: So IsEncryptedError can check the error code by checking 0 < x < 5 instead of matching random values

cosmwasm/packages/wasmi-runtime/src/registration/cert.rs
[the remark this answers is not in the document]
reuvenpo: ahhhh i see. then we can just remove it, or replace it with a comment, i think

cosmwasm/packages/wasmi-runtime/src/registration/attestation.rs
```diff
@@ -535,7 +552,7 @@ pub fn make_ias_client_config() -> rustls::ClientConfig {
 
 #[cfg(feature = "SGX_MODE_HW")]
 pub fn get_sigrl_from_intel(fd: c_int, gid: u32) -> Vec<u8> {
-    info!("get_sigrl_from_intel fd = {:?}", fd);
+    trace!("get_sigrl_from_intel fd = {:?}", fd);
     let config = make_ias_client_config();
     let ias_key = String::from_utf8_lossy(API_KEY_FILE).trim_end().to_owned();
```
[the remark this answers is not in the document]
Cashmaney: Very offtopic, yes. But the reason is that there are some things that we are fine with exploding if they go wrong - specifically, stuff that is run during the offline phase of the registration. That's I created the 2 different files - offchain.rs and onchain.rs. Online is the stuff run onchain that needs to be graceful, and offchain contains a lot of communication and parsing that we don't care about handling if it fails

x/compute/internal/types/errors.go
reuvenpo: ahhhh, cool cool cool. I'd add a comment about it in this section of the code just to avoid subtle bugs if anyone adds to this list.

cosmwasm/packages/wasmi-runtime/src/registration/attestation.rs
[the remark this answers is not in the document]
reuvenpo: ahhh, i see. i suppose that makes sense

x/compute/internal/types/errors.go
reuvenpo: i see... if you think that it's safe and sound, then we can resolve this conversation.

PR #397, opened 2020-07-13
Node auth error messages - take 3

### LINKED ISSUE #376: Return attestation verification (on chain) error
Error in attestation isn't returned as a log, so it's hard to know what went wrong beyond running a node and checking the log, which kind of defeats the purpose

reuvenpo:
This PR makes errors in node authentication display more useful errors for users that encounter them.
closes #376 
To reviewer: The technique shown here to generate nicer error messages can and should be applied to `EnclaveError`.


go-cosmwasm/src/error/rust.rs
```diff
@@ -35,6 +35,12 @@ pub enum Error {
         #[cfg(feature = "backtraces")]
         backtrace: snafu::Backtrace,
     },
+    #[snafu(display("Error calling the enclave: {}", msg))]
```
Cashmaney, line 38: This isn't an error _calling_ the enclave specifically. Also if this is going to be printed to the log we can probably just print `"{}", msg` since it will be an informative error

cosmwasm/packages/enclave-ffi-types/src/types.rs
```diff
@@ -92,6 +92,28 @@ pub enum EnclaveError {
     Unknown,
 }
 
+/// This type represents the possible error conditions that can be encountered in the
+/// enclave while authenticating a new node in the network.
+/// cbindgen:prefix-with-name
+#[repr(C)]
+#[derive(Debug, Display, PartialEq, Eq)]
+pub enum NodeAuthResult {
```
Cashmaney, line 100: Do you think we could add errors for specific isvEnclaveQuoteStatus? Like, return "failed to perform registration. Invalid quote status "KEY_REVOKED" " or something

go-cosmwasm/src/error/rust.rs
Cashmaney: Also this is the 3rd variant of the name `EnclaveErr` in the project. Maybe name this something else? (sgx-vm `EnclaveErr` and `EnclaveError` in ffi/types)

cosmwasm/packages/enclave-ffi-types/src/types.rs
reuvenpo: hmm... can you point me to the place in the code where this error originates? I'd probably have to fiddle with 1-3 functions to bubble this specific case out to the context where i instantiate variants of this type.
Note that the way i use this type is really high level, and unrelated to how most of the related code just uses `sgx_status_t`

go-cosmwasm/src/error/rust.rs
reuvenpo: Ok, so just writing is as `"{}", msg` is probably alright.
I was debating creating another variant `SdkError` for cases of `sgx_status_t` like i did in `VmError`, for which this message is accurate, and then change the message on this variant to something like "Error returned from the enclave", but i felt that might be excessive.
As for the naming, the name shows up in a lot of contexts, but that's because different contexts use different error types, and all of them needed a new variant for errors that come from the enclave. I agree that the naming duplication isn't great, but for each of those contexts separately i felt like that was the most appropriate name.
In any case, if you mix them up, the compiler will let you know.

cosmwasm/packages/enclave-ffi-types/src/types.rs
Cashmaney: https://github.com/enigmampc/SecretNetwork/blob/05787bae238375af0e95662e3006ba56451f9590/cosmwasm/packages/wasmi-runtime/src/registration/cert.rs#L360

If it's not something you can do in 15 minutes or less we can leave it for some other time

go-cosmwasm/src/error/rust.rs
Cashmaney: So why not call it GoEnclaveErr or something? Duplicate names will cause all sorts of fun when trying to use find and replace, or find in path (not to mention the potential `Expected EnclaveErr but got EnclaveErr`  good times)

reuvenpo: It's not even in Go yet :sweat_smile: 
But we can name it `GoCwEnclaveError` or something to signal the layer it's in.

Cashmaney: יאללה קניתי

assafmo: image

cosmwasm/packages/enclave-ffi-types/src/types.rs
reuvenpo: Yeah, this will be somewhat involved.
I'm not sure I understand what exactly all the structs and enums mean, and the specific error you're referring to happens a couple layers into functions that obfuscate the specific errors that happen. If I change that, I'd need to add a couple of error types, and modify a bunch of other functions that use the same functions that I needed to change in the first place.

PR #417, opened 2020-07-21
migrate-testnet-with-sgx

mabalaru:
I created the document for validator migration on testnet with SGX and unjail commands.


mabalaru:
Hey @assafmo 
there are 2 differences: the doc used to bring up the full node (on testnet you have to use the one with SGX and u have to follow the sgx steps to register the new node) + the unjail.
if you consider it unnecessarily for testnet you can close the PR.

thanks

assafmo:
I think it's not necessary for now. Thank you anyway!

PR #419, opened 2020-07-22
New and improved input validation

Cashmaney:
Init msg now must include the code hash encrypted with the tx

Exec msg now must include the expected contract key encrypted with the tx

Queries unchanged for now

Contract key is now stored in the contract db on init


x/compute/internal/keeper/keeper_test.go
```diff
@@ -480,12 +480,18 @@ func TestExecuteWithCpuLoop(t *testing.T) {
...
+	defer func() {
+		r := recover()
+		require.NotNil(t, r)
+		_, ok := r.(sdk.ErrorOutOfGas)
+		require.True(t, ok, "%v", r)
+	}()
+
 	// this must fail
 	_, err = keeper.Execute(ctx, addr, fred, execMsgBz, nil)
 	assert.Error(t, err)
```
assafmo, line 493: Please make this `require.True(t, false)`.

x/compute/internal/keeper/secret_contracts_test.go
```diff
@@ -1040,19 +1041,22 @@ func TestAllocateOnHeapFailBecauseGasLimit(t *testing.T) {
...
+	_, _, _ = execHelper(t, keeper, ctx, addr, walletA, `{"allocate_on_heap":{"bytes":1073741824}}`, false, defaultGas)
 
 	// this should fail with out of gas because 1GiB will ask for
 	// 134,217,728 gas units (8192 per page times 16,384 pages)
 	// the default gas limit in ctx is 200,000 which translates into
 	// 20,000,000 WASM gas units, so before the memory_grow opcode is reached
 	// the gas metering sees a request that'll cost 134mn and the limit
 	// is 20mn, so it throws an out of gas exception
-
-	require.Empty(t, data)
```
assafmo, line 1002: Please add here `require.True(t, false)`.

x/compute/client/cli/query.go
```diff
@@ -195,6 +220,44 @@ func GetCmdGetContractInfo(cdc *codec.Codec) *cobra.Command {
...
+
+			encodedInput := args[0]
+
+			dataCipherBz, err := base64.StdEncoding.DecodeString(encodedInput)
+			if err != nil {
+				return fmt.Errorf("error while trying to decode the encrypted output data from base64: %w", err)
+			}
+
+			nonce := dataCipherBz[0:32]
+			originalTxSenderPubkey := dataCipherBz[32:64]
```
reuvenpo, line 242: This feels very fragile to me if we change the schema of these messages. Do we have a canonical place in the code the parses these?

x/compute/client/cli/tx.go
```diff
@@ -135,16 +139,22 @@ func InstantiateContractCmd(cdc *codec.Codec) *cobra.Command {
...
+			initMsg.CodeHash, err = GetCodeHashByCodeId(cliCtx, args[0])
+			if err != nil {
+				return err
+			}
 
-			initMsg := []byte(args[1])
-			initMsg, err = wasmCtx.Encrypt(initMsg)
+			// todo: Add check that this is valid json and stuff
+			initMsg.Msg = []byte(args[1])
+
```
reuvenpo, line 156: You don't really need to check that it's a JSON, since the format is defined in the contract

x/compute/client/cli/tx.go
```diff
@@ -187,9 +197,11 @@ func ExecuteContractCmd(cdc *codec.Codec) *cobra.Command {
 			inBuf := bufio.NewReader(cmd.InOrStdin())
 			txBldr := auth.NewTxBuilderFromCLI(inBuf).WithTxEncoder(utils.GetTxEncoder(cdc))
 			cliCtx := context.NewCLIContextWithInput(inBuf).WithCodec(cdc)
+			wasmCtx := wasmUtils.WASMContext{CLIContext: cliCtx}
+			execMsg := types.SecretMsg{}
+			contractAddr := sdk.AccAddress{}
 
-			var contractAddr = sdk.AccAddress{}
-			var execMsg []byte
+			// var execMsg []byte
```
reuvenpo, line 204: remove this comment

x/compute/internal/keeper/keeper_test.go
```diff
@@ -237,7 +246,8 @@ func TestInstantiate(t *testing.T) {
 	require.Equal(t, "secret18vd8fpwxzck93qlwghaj6arh4p7c5n8978vsyg", addr.String())
 
 	gasAfter := ctx.GasMeter().GasConsumed()
-	require.Equal(t, uint64(37733), gasAfter-gasBefore)
+	require.Greater(t, gasAfter-gasBefore, uint64(20000))
+	require.Less(t, gasAfter-gasBefore, uint64(60000))
```
reuvenpo: We usually put it exact numbers here to make sure that the code is deterministic.

x/compute/internal/keeper/keeper_test.go
```diff
@@ -447,26 +452,47 @@ func TestExecuteWithCpuLoop(t *testing.T) {
...
 	require.NoError(t, err)
 
+	// ensure we get an out of gas panic
+	//defer func() {
+	//	r := recover()
+	//	require.NotNil(t, r)
+	//	_, ok := r.(sdk.ErrorOutOfGas)
+	//	require.True(t, ok, "%v", r)
+	//}()
+
```
reuvenpo: why is this comment here?
I expect that a panic should be thrown, no?

x/compute/internal/keeper/secret_contracts_test.go
```diff
@@ -185,7 +209,15 @@ func execHelper(t *testing.T, keeper Keeper, ctx sdk.Context, contractAddress sd
...
-	initMsgBz, err := wasmCtx.Encrypt([]byte(initMsg))
+
+	hash := keeper.GetCodeInfo(ctx, codeID).CodeHash
+
+	msg := types.SecretMsg{
+		CodeHash: []byte(hex.EncodeToString(hash)),
+		Msg:      []byte(initMsg),
+	}
+
+	initMsgBz, err := wasmCtx.Encrypt(msg.Serialize())
```
reuvenpo, line 220: This pattern repeats a lot. can it maybe be a helper function? or a part of `wasmCtx.Encrypt`?

cosmwasm/packages/enclave-ffi-types/src/types.rs
```diff
@@ -58,7 +58,8 @@ pub enum EnclaveError {
     // TODO should we split these three cases for better diagnostics?
     #[display(fmt = "failed to execute ocall")]
     FailedOcall { vm_error: UntrustedVmError },
-
+    #[display(fmt = "failed to validate transaction")]
+    ValidationFailure,
```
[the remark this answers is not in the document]
Cashmaney: This is about validating the transaction, not only the inputs. The PR is called input validation because this originally started as a way to bind the user input to a contract, but it's no longer just about that. Also, you want to keep these error messages in the language a user will understand. "failed to validate transaction" is a good way to tell him he fucked up somewhere IMO, but I'm open to suggestions

cosmwasm/packages/enclave-ffi-types/src/types.rs
```diff
@@ -73,7 +74,8 @@ pub enum EnclaveError {
     /// Fail to inject gas metering
     #[display(fmt = "failed to inject gas metering")]
     FailedGasMeteringInjection,
-
+    #[display(fmt = "internal error during execution")]
+    InternalError,
```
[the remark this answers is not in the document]
Cashmaney: The point was an error code similar to http error code 500. The error isn't `unknown`, it just doesn't matter to the user what happened there. Though I'm pretty sure I didn't end up using it in the end, I still like the distinction from unknown

cosmwasm/packages/wasmi-runtime/src/wasm/contract_operations.rs
```diff
@@ -53,12 +53,12 @@ pub fn init(
         EnclaveError::FailedToDeserialize
     })?;
 
-    let contract_key = generate_encryption_key(&parsed_env, contract)?;
+    let contract_address = &parsed_env.contract.address;
+    let contract_key = generate_encryption_key(&parsed_env, contract, contract_address.as_slice())?;
```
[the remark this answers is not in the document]
Cashmaney: The original point was to bind the contract address to the contract key, which was useful in the first iteration when we were checking the contract key, not the code hash. In this context it isn't forgeable, since offline attacks are not relevant to this use.

It isn't important anymore, but I left it there because it was already written, and it doesn't hurt like you said

cosmwasm/packages/wasmi-runtime/src/wasm/contract_validation.rs
```diff
@@ -99,14 +108,39 @@ pub fn validate_contract_key(
...
+        return Err(EnclaveError::ValidationFailure);
+    }
+
+    let contract_hash = calc_contract_hash(contract_code);
+
+    let mut encrypted_contract_hash: [u8; HEX_ENCODED_HASH_SIZE] = [0u8; HEX_ENCODED_HASH_SIZE];
+    encrypted_contract_hash.copy_from_slice(&msg[0..HEX_ENCODED_HASH_SIZE]);
+
+    let decoded_hash: Vec<u8> =
+        hex::decode(encrypted_contract_hash.to_vec().as_slice()).map_err(|_| {
```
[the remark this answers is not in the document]
Cashmaney: People will actually need to read and write these values to call transactions (for example, you now have to hard code the target contract address and code hash). They're encoded to make them easy to query from the CLI and paste into your code

x/compute/client/cli/tx.go
Cashmaney: You don't _need_ to, but if you make a mistake it gives you a weird error instead of telling you exactly what you did wrong

x/compute/internal/keeper/keeper_test.go
Cashmaney: If you want to test determinism, it should be in a test called `testDeterminism` with more specific stuff. This way it just annoys anyone who makes any change to instantiate since you get a failure in `testInstantiate` for no apparent reason

cosmwasm/packages/wasmi-runtime/src/wasm/contract_validation.rs
```diff
@@ -99,14 +108,39 @@ pub fn validate_contract_key(
...
     let calculated_authentication_id =
-        generate_contract_id(&enclave_key, &signer_id, &contract_hash);
+        generate_contract_id(&enclave_key, &signer_id, &contract_hash, contract_address);
 
     calculated_authentication_id == expected_authentication_id
 }
+
+pub fn validate_msg(msg: &[u8], contract_code: &[u8]) -> Result<Vec<u8>, EnclaveError> {
+    if msg.len() < HEX_ENCODED_HASH_SIZE {
+        error!("Got message that is missing contract hash for validation");
```
toml01, line 125: I think a better message would be "Malformed message" or something similar? This message could be misleading

x/compute/client/cli/tx.go
reuvenpo: still, if you add this though, you'll lock down contract from using other formats if they choose to use something different than the current `cosmwasm-std`

cosmwasm/packages/wasmi-runtime/src/wasm/contract_validation.rs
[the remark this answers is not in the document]
reuvenpo: Ah I see, this makes sense

x/compute/client/cli/tx.go
Cashmaney: It can come with a flag that turns the check off.. it's just a CLI thing, and I'd rather people have an easier time to use what we have now than have an easier time supporting some advanced feature that we don't know anyone will want

reuvenpo: I suppose that's fair. if we need to we can loosed that in the future. The messages that the contract returns are required to be JSON anyways, so it's unlikely people will do anything different here.

cosmwasm/packages/wasmi-runtime/src/wasm/query_chain.rs
```diff
@@ -230,11 +230,14 @@ fn encrypt_query_request(
     let mut is_encrypted = false;
 
     // encrypt message
-    if let QueryRequest::Wasm(WasmQuery::Smart { msg, .. }) = query_struct {
+    if let QueryRequest::Wasm(WasmQuery::Smart { msg, callback_code_hash, .. }) = query_struct {
         is_encrypted = true;
 
+        let mut hash_appended_msg = callback_code_hash.clone().as_bytes().to_vec();
+        hash_appended_msg.extend_from_slice(&msg.0.clone());
```
reuvenpo, line 237: Save two unnecessary copies by writing:
```rust
let mut hash_appended_msg = callback_code_hash.clone().into_bytes();
hash_appended_msg.extend_from_slice(&msg.0); // or maybe &msg.0[..]
```

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
```diff
@@ -123,12 +123,27 @@ fn encrypt_wasm_msg(
...
+
+            let mut msg_to_pass = SecretMessage::from_base64(
+                Binary(hash_appended_msg).to_base64(),
+                nonce,
+                user_public_key,
+            )?;
 
             msg_to_pass.encrypt_in_place()?;
-            *msg = b64_encode(&msg_to_pass.to_slice());
+            *msg = Binary::from(msg_to_pass.to_vec().as_slice());
```
reuvenpo, line 146: replace with
```rust
*msg = Binary(msg_to_pass.to_vec());
```
to make one less copy

cosmwasm/packages/wasmi-runtime/src/wasm/types.rs
```diff
@@ -95,7 +95,7 @@ impl SecretMessage {
         })
     }
 
-    pub fn to_slice(&self) -> Vec<u8> {
+    pub fn to_vec(&self) -> Vec<u8> {
```
reuvenpo, line 98: great change :D

go-cosmwasm/api/lib_test.go
```diff
@@ -96,6 +96,7 @@ func binaryAddr(human string) []byte {
 }
 
 func TestInstantiate(t *testing.T) {
+	t.SkipNow()
 	cache, cleanup := withCache(t)
 	defer cleanup()
```
reuvenpo, line 102: why are all these tests skipped?

cosmwasm/testing/callback-test.js
```diff
@@ -43,8 +48,8 @@ const assert = require("assert").strict;
     }
   );
 
-  const execTx = await signingClient.execute(contract, {
-    a: { contract_addr: contract, x: 2, y: 3 },
+  const execTx = await signingClient.execute(contractAddr, {
+    a: { contract_addr: contractAddr, code_hash: contractCodeHash, x: 2, y: 3 },
```
reuvenpo, line 52: how does this work if it's not called `code_hash`?

go-cosmwasm/types/msg.go
```diff
@@ -94,6 +94,9 @@ type ExecuteMsg struct {
 	// ContractAddr is the sdk.AccAddress of the contract, which uniquely defines
 	// the contract ID and instance ID. The sdk module should maintain a reverse lookup table.
 	ContractAddr string `json:"contract_addr"`
+	// Custom addition to support binding a message to specific code to harden against offline & replay attacks
+	// This is only needed when creating a callback message
+	CallbackCodeHash string `json:"callback_code_hash"`
 	// Msg is assumed to be a json-encoded message, which will be passed directly
```
reuvenpo, line 100: Also change these definitions

go-cosmwasm/types/msg.go
```diff
@@ -104,6 +107,9 @@ type ExecuteMsg struct {
 type InstantiateMsg struct {
 	// CodeID is the reference to the wasm byte code as used by the Cosmos-SDK
 	CodeID uint64 `json:"code_id"`
+	// Custom addition to support binding a message to specific code to harden against offline & replay attacks
+	// This is only needed when creating a callback message
+	CallbackCodeHash string `json:"callback_code_hash"`
 	// Msg is assumed to be a json-encoded message, which will be passed directly
```
reuvenpo, line 113: and this

go-cosmwasm/types/types_test.go
```diff
@@ -1,27 +0,0 @@
...
-func TestCanoncialAddress(t *testing.T) {
-	addr := CanonicalAddress([]byte{17, 121, 0, 93})
-	bz, err := json.Marshal(addr)
-	if err != nil {
-		t.Error(err)
-	}
-	if string(bz) != "[17,121,0,93]" {
-		t.Errorf("Unexpected encoding %s", string(bz))
-	}
-
```
reuvenpo, line 18: I'd say don't remove this file, because if they change it upstream it's just going to give us conflicts.

x/compute/client/cli/tx.go
```diff
@@ -176,18 +192,20 @@ func InstantiateContractCmd(cdc *codec.Codec) *cobra.Command {
...
 			// build and sign the transaction, then broadcast to Tendermint
 			msg := types.MsgInstantiateContract{
-				Sender:    cliCtx.GetFromAddress(),
-				Code:      codeID,
-				Label:     label,
-				InitFunds: amount,
-				InitMsg:   initMsg,
-				Admin:     adminAddr,
+				Sender:           cliCtx.GetFromAddress(),
+				CallbackCodeHash: "",
```
reuvenpo, line 196: I probably don't understand this context completely, but - why is `CallbackCodeHash` here empty?

x/compute/client/cli/tx.go
```diff
@@ -244,36 +263,76 @@ func ExecuteContractCmd(cdc *codec.Codec) *cobra.Command {
...
 
 			// build and sign the transaction, then broadcast to Tendermint
 			msg := types.MsgExecuteContract{
-				Sender:    cliCtx.GetFromAddress(),
-				Contract:  contractAddr,
-				SentFunds: amount,
-				Msg:       execMsg,
+				Sender:           cliCtx.GetFromAddress(),
+				Contract:         contractAddr,
+				CallbackCodeHash: "",
```
reuvenpo, line 297: same question here - why is this empty?

x/compute/client/rest/query.go
```diff
@@ -185,3 +187,69 @@ func (a *argumentDecoder) DecodeString(s string) ([]byte, error) {
 		return a.dec(s)
 	}
 }
+
+func queryCodeHashHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		codeID, err := strconv.ParseUint(mux.Vars(r)["codeID"], 10, 64)
+		if err != nil {
+			rest.WriteErrorResponse(w, http.StatusInternalServerError, err.Error())
```
reuvenpo, line 195: Sounds like it should be a 400 Bad Request

x/compute/client/rest/query.go
```diff
@@ -185,3 +187,69 @@ func (a *argumentDecoder) DecodeString(s string) ([]byte, error) {
...
+	}
+}
+
+func queryContractHashHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		contractAddr, err := sdk.AccAddressFromBech32(mux.Vars(r)["contractAddr"])
+		if err != nil {
+			rest.WriteErrorResponse(w, http.StatusInternalServerError, err.Error())
+			return
+		}
```
reuvenpo, line 234: also, this is a 400 error

x/compute/client/rest/tx.go
```diff
@@ -113,12 +117,18 @@ func instantiateContractHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
 			return
 		}
 
+		//res, err := GetCodeHashByCodeId(cliCtx, string(codeID))
+		//if err != nil {
+		//	return
+		//}
+
```
reuvenpo, line 124: remove this kind of comments you added to this files

x/compute/client/rest/tx.go
```diff
@@ -113,12 +117,18 @@ func instantiateContractHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
...
 		msg := types.MsgInstantiateContract{
-			Sender:    cliCtx.GetFromAddress(),
-			Code:      codeID,
-			InitFunds: req.Deposit,
-			InitMsg:   req.InitMsg,
-			Admin:     req.Admin,
+			Sender:            cliCtx.GetFromAddress(),
+			Code:      		   codeID,
+			CallbackCodeHash:  "",
```
reuvenpo, line 128: again, why is this empty?

x/compute/client/rest/tx.go
```diff
@@ -166,3 +182,30 @@ func executeContractHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
...
+
+func GetCodeHashByContractAddr(cliCtx context.CLIContext, contractAddr sdk.AccAddress) ([]byte, error) {
+	route := fmt.Sprintf("custom/%s/%s/%s", types.QuerierRoute, keeper.QueryContractHash, contractAddr.String())
+	res, _, err := cliCtx.Query(route)
+	if err != nil {
+		return nil, err
+	}
+
+	return []byte(hex.EncodeToString(res)), nil
+}
```
reuvenpo, line 194: feels like this pattern keeps repeating. is it not close enough to wrap up in a function to prevent the duplication?

cosmwasm/contracts/staking/Makefile
```diff
@@ -0,0 +1,9 @@
+all:
+	RUSTFLAGS='-C link-arg=-s' cargo build --release --target wasm32-unknown-unknown --locked
+	## The following line is not necessary, may work only on linux (extra size optimization)
+	wasm-opt -Os ./target/wasm32-unknown-unknown/release/*.wasm -o ./contract.wasm
+	cat ./contract.wasm | gzip -9 > ./contract.wasm.gz
+
+clean:
+	cargo clean
+	-rm -f ./contract.wasm ./contract.wasm.gz
```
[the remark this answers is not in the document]
Cashmaney: Why not keep it? Otherwise it's annoying to recompile the contract

cosmwasm-js/packages/sdk/src/types.ts
```diff
@@ -91,6 +91,7 @@ export interface MsgExecuteContract extends MsgTemplate {
     /** Bech32 account address */
     readonly contract: string;
     /** Handle message as JavaScript object */
+    readonly callback_code_hash: string;
     msg: any;
```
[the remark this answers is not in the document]
Cashmaney: I changed it to `callback_code_hash` to make it clearer that the usage is for when you're creating a callback message, not for regular messages. If it just `code_hash` then it won't be as clear what it's used for and when you can keep it blank. I'm open to suggestions that make more intuitive

x/compute/client/rest/tx.go
Cashmaney: Because this is just a regular message and not a callback, so the code hash is already contained in the encrypted message

x/compute/client/rest/tx.go
Cashmaney: How would you do it differently?

cosmwasm/contracts/staking/Makefile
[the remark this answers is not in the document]
reuvenpo: that's fair

x/compute/client/rest/tx.go
reuvenpo: I'd create a wrapper for `context.CLIContext` (I understand that it comes from another package?) and then implement these queries as methods on it. that way you have an obvious place to write all the different queries you need to do, and you get one thing that exposes all available queries in a structured way.

Cashmaney: Ah, that could work. You'd have to create the new type from CLIContext every time you wanted to use the functions though, but if there are a few of them it might be worth it. In this case I deleted this since it was unnecessary anyway, but we can keep this is mind if we ever need to add more

reuvenpo: :+1:

go-cosmwasm/api/lib_test.go
assafmo: These are cosmwasm tests that we never ran and want to start running, but they're calling wasm without encryption. I'm not sure whether we need to fix them to leave it like that.
There are other tests in this module that do pass.

reuvenpo: aha, that makes sense :+1:

PR #421, opened 2020-07-24
Testnet docs: Clarification of CLI, faucet, create validator

AronVanAmmers:
- Explain how the CLI can be run on a different machine
- Add instructions for getting test-SCRT from the faucet
- Add last step to create a validator. In the mainnet docs there is a separate MD file for this, but I think having it as a step here makes it a nice concise complete instruction.


Cashmaney:
Thanks @AronVanAmmers !

Issue #424, opened 2020-07-25
Add a secretd command to check if enclave is working


Cashmaney:
We should add a command that lets people check if enclave is working properly (health check SGX). Would go a long way to making troubleshooting easier

PR #428, opened 2020-07-26
#424 secretd enclave working

Cashmaney:
#424 Added a small health check that prints out the enclave status

### DIFF
--- a/cosmwasm/packages/enclave-ffi-types/src/types.rs
+++ b/cosmwasm/packages/enclave-ffi-types/src/types.rs
@@ -165,6 +165,21 @@ pub enum NodeAuthResult {
     Panic,
 }
 
+/// This type represents the possible error conditions that can be encountered in the
+/// enclave while authenticating a new node in the network.
+/// cbindgen:prefix-with-name
+#[repr(C)]
+#[derive(Debug, Display, PartialEq, Eq)]
+pub enum HealthCheckResult {
+    Success,
+}
+
+impl Default for HealthCheckResult {
+    fn default() -> Self {
+        HealthCheckResult::Success
+    }
+}
+
 /// This type holds a pointer to a VmError that is boxed on the untrusted side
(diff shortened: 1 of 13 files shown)


cosmwasm/packages/enclave-ffi-types/src/types.rs
```diff
@@ -165,6 +165,21 @@ pub enum NodeAuthResult {
...
 }
 
+/// This type represents the possible error conditions that can be encountered in the
+/// enclave while authenticating a new node in the network.
+/// cbindgen:prefix-with-name
+#[repr(C)]
+#[derive(Debug, Display, PartialEq, Eq)]
+pub enum HealthCheckResult {
+    Success,
+}
```
reuvenpo, line 175: This enum only has one variant? that's odd

cosmwasm/packages/wasmi-runtime/src/exports.rs
```diff
@@ -181,3 +181,11 @@ pub unsafe extern "C" fn ecall_query(
...
     }
 }
+
+/// # Safety
+/// Always use protection
+#[no_mangle]
+pub unsafe extern "C" fn ecall_health_check(
+) -> HealthCheckResult {
+    HealthCheckResult::Success
+}
```
reuvenpo, line 191: Is this intentionally empty? You could add some checks here, theoretically.
also, why the newline between the function definition's `()`s?

cosmwasm/packages/enclave-ffi-types/src/types.rs
Cashmaney: One day we might want to make this check other things and return more detailed responses. So it's an enum for easy upgradability

cosmwasm/packages/wasmi-runtime/src/exports.rs
Cashmaney: Intentionally empty, yes. When we want to check other things we can add to it, but right now I wanted something that just reports if the enclave is running fine or not

PR #431, opened 2020-07-26
Enclave tests

reuvenpo:
This PR adds support for writing tests in wasmi-runtime that will run inside a real SGX enclave.
I add two make commands:
- from the project directory: `make enclave-tests` - runs the test suite
- from the `cosmwasm/packages/enclave-test` directory: `make check` to run the `cargo check` on the test code.

I laid things out as follows:
- An ecall called `ecall_run_tests` is always exported from the enclave.
    - When compiled with the `test` flag, this ecall will run the test suite. When compiled without it, it does nothing
    - functions called by this mechanism should be annotated with `#[cfg(feature = "test")]` or be placed in a module annotated with it.
- This ecall just calls all the tests that it wants to run, one after the other.
- I show a couple examples of how submodules can help organize the test code
- `cosmwasm-sgx-vm` exports a wrapper for the ecall when compiled with the `enclave-tests` flag
- `enclave-test` calls this wrapper

The test runner inside the enclave runs each test in sequence, counting the number of failures and reporting the status per test.
It then returns the number of failed tests to the caller. `enclave-test` will return a nonzero status code when either calling the enclave fails, or the enclave reports that a test has failed.


cosmwasm/packages/sgx-vm/src/enclave_tests.rs
```diff
@@ -0,0 +1,10 @@
+// I reexport `sgx_types` here so that `go-cosmwasm` doesn't need to specify it as a dev-dependency.
+pub use sgx_types;
+use sgx_types::{sgx_enclave_id_t, sgx_status_t};
+
+pub use crate::enclave::get_enclave;
+
+#[link(name = "rust_cosmwasm_enclave.signed")]
+extern "C" {
+    pub fn ecall_run_tests(eid: sgx_enclave_id_t) -> sgx_status_t;
```
Cashmaney, line 9: To be able to run this in CI we would need this to return an exit code. Could you make it so the function returns an integer or something so we have a success/fail indicator?

cosmwasm/packages/sgx-vm/src/enclave_tests.rs
```diff
@@ -0,0 +1,10 @@
+// I reexport `sgx_types` here so that `go-cosmwasm` doesn't need to specify it as a dev-dependency.
```
Cashmaney, line 1: Can we move this file to a tests folder? Same as you did in go-cosmwasm

cosmwasm/packages/sgx-vm/src/enclave_tests.rs
```diff
@@ -0,0 +1,10 @@
...
+pub use sgx_types;
+use sgx_types::{sgx_enclave_id_t, sgx_status_t};
+
+pub use crate::enclave::get_enclave;
+
+#[link(name = "rust_cosmwasm_enclave.signed")]
```
Cashmaney, line 7: Don't forget to remove this?

cosmwasm/packages/sgx-vm/src/enclave_tests.rs
reuvenpo: I'd ideally like to return a full test report. So, per test whether it succeeded or failed (panicked).
That would require an ocall or something though, since it'll have a dynamic size and i don't like that...
How about returning the number of failed tests, so if no tests fail, we'll get the nice "0" which is also a success status?

reuvenpo:
>That way you avoid the long compilation times, dev dependency stuff, and it would be more straightforward to use as a binary that returns an exit code on success/failure

I like this Idea. we really don't need go-cosmwasm here. I'll play around with this idea

cosmwasm/packages/sgx-vm/src/enclave_tests.rs
Cashmaney: Yep, returning the amount of failures should be fine. That's usually what test runners do afaik

PR #435, opened 2020-07-29
Update azuredeploy.json

Cashmaney:


packaging_docker/testnet/azure/secret-node-azure-template/azuredeploy.json
```diff
@@ -127,7 +127,7 @@
       "type": "Microsoft.Network/networkSecurityGroups",
       "apiVersion": "2020-04-01",
       "name": "[variables('networkSecurityGroupName')]",
-      "location": "uksouth",
+      "location": "[parameters('location')]",
```
reuvenpo, line 130: where does this come from?

packaging_docker/testnet/node_init.sh
```diff
@@ -28,6 +28,12 @@ then
 
   sed -i 's/persistent_peers = ""/persistent_peers = "'"$PERSISTENT_PEERS"'"/g' ~/.secretd/config/config.toml
   echo "Set persistent_peers: $PERSISTENT_PEERS"
+  
+  # Open RPC port to all interfaces
+  perl -i -pe 's/laddr = .+?26657"/laddr = "tcp:\/\/0.0.0.0:26657"/' ~/.secretd/config/config.toml
+
+  # Open P2P port to all interfaces
+  perl -i -pe 's/laddr = .+?26656"/laddr = "tcp:\/\/0.0.0.0:26656"/' ~/.secretd/config/config.toml
```
reuvenpo, line 36: I don't care either way, but looks like other edits in this file were written with `sed`. It should probably be consistent (and again, i don't care which command you use here)

Cashmaney: Copied these from the azure one, and I'd rather keep them as close as possible

packaging_docker/testnet/azure/secret-node-azure-template/azuredeploy.json
Cashmaney: The parameters field in the top level. Those are the values you accept from the user etc.

packaging_docker/testnet/node_init.sh
reuvenpo: fair enough

PR #437, opened 2020-07-30
Make query gas limit configurable for nodes

assafmo:
- `secred init`: append `query-gas-limit = 50000` to `.secretd/config/app.toml`
- `secretd start`: read `query-gas-limit` from `.secretd/config/app.toml` and pass it to x/compute instead of the default 30k value

### DIFF
--- a/cmd/secretd/main.go
+++ b/cmd/secretd/main.go
@@ -94,7 +94,7 @@ func main() {
 # query-gas-limit sets the gas limit under which your node will run smart sontracts queries.
 # Queries that consume more than this value will be terminated prematurely with an error.
 # This is a good way to protect your node from DoS by heavy queries.
-query-gas-limit = 50000
+query-gas-limit = 3000000
 `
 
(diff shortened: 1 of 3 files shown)


cmd/secretd/main.go
```diff
@@ -80,6 +83,26 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
+
+	// Set default gas limit for WASM queries
+	queryGasLimitTemplate := `
```
toml01, line 88: You should edit this:
https://github.com/cosmos/cosmos-sdk/blob/v0.38.3/server/config/toml.go

Instead of putting it here hardcoded

assafmo: It's inside the sdk

toml01: Yep, don't we use our own fork for a lot of things?

assafmo: Yeah, but I think we only fixed the multisig issue and upgraded v0.38.3 to use Tendermint v0.33.6.
The second link you sent is synced with their master which is somewhere between v0.39 to v0.40.

Essentially this is what we have right now on mainnet: https://github.com/cosmos/cosmos-sdk/compare/v0.38.3...enigmampc:v0.38.3-scrt-0.2.1

PR #445, opened 2020-08-03
Refactor output encryption

### LINKED ISSUE #306: Improve JSON parsing while encrypting outputs
The JSON usage in `cosmwasm/packages/wasmi-runtime/src/wasm/io.rs` while encrypting the outputs can probably be simplified.

toml01:
fixes #306 
Code is now much less fragile and much more readable.


cosmwasm/packages/wasmi-runtime/src/cosmwasm/types.rs
```diff
@@ -104,27 +105,130 @@ pub struct Coin {
...
+    },
+    OkString {
+        #[serde(rename = "Ok")]
+        ok: String,
+    },
+    OkObject {
+        #[serde(rename = "Ok")]
+        ok: OkObject,
+    },
+}
```
reuvenpo, line 122: You can rewrite this as:
```rust
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
pub enum WasmOutput {
  Ok(OkObjectOrString),
  Err(Value)
}

#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
#[serde(untagged)]
pub enum OkObjectOrString {
  Str(String),
  Obj(OkObject),
}
```
Which should generate the exact same parser.
Then again the way you did it makes the match later nicer...
So go with what you think is better - after writing this i like your way better, just wanted to show you you could simplify the definition here.

cosmwasm/packages/wasmi-runtime/src/cosmwasm/types.rs
```diff
@@ -104,27 +105,130 @@ pub struct Coin {
...
+where
+    T: Clone + fmt::Debug + PartialEq,
+{
+    Bank(BankMsg),
+    // by default we use RawMsg, but a contract can override that
+    // to call into more app-specific code (whatever they define)
+    Custom(T),
+    Staking(StakingMsg),
+    Wasm(WasmMsg),
+}
```
reuvenpo, line 151: Seems like you changed the definition of `CosmosMsg`. why is that?
After another look - did you basically refresh an old definition with how it looks like upstream now?

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
```diff
@@ -65,99 +66,89 @@ pub fn encrypt_output(
...
+            // Encrypting the actual error
+            let encrypted_err = encrypt_serializeable(&key, &err)?;
+
+            // Creating a 'generic_err' envelope
             let mut new_value: Value = json!({"generic_err":{"msg":""}});
-            new_value["generic_err"]["msg"] = encrypt_serializeable(&key, &err)?;
-            v["Err"] = new_value;
+            new_value["generic_err"]["msg"] = Value::String(encrypted_err);
+
+            *err = new_value;
```
reuvenpo, line 87:
```rust
WasmOutput::ErrObject { err } => {
  let encrypted_err = encrypt_serializeable(&key, err)?;  // note no & before `err`
  *err = json!({"generic_err": {"msg": encrypted_err}});
}
```

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
```diff
@@ -65,99 +66,89 @@ pub fn encrypt_output(
...
+
+fn encrypt_wasm_msg(
+    wasm_msg: &mut WasmMsg,
+    nonce: IoNonce,
+    user_public_key: Ed25519PublicKey,
+) -> Result<(), EnclaveError> {
+    match wasm_msg {
+        WasmMsg::Execute { msg, .. } => {
+            let mut msg_to_pass =
+                SecretMessage::from_base64((*msg).to_string(), nonce, user_public_key)?;
```
reuvenpo, line 139: `(*msg).to_string()` is just `msg.clone()`

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
```diff
@@ -65,99 +66,89 @@ pub fn encrypt_output(
...
-                                e["value"] = encrypt_serializeable(&key, v)?;
-                            }
-                        }
-                    }
-                }
+            // Encrypt all logs
+            for log in &mut ok.log {
+                log.key = encrypt_serializeable(&key, &log.key)?;
+                log.value = encrypt_serializeable(&key, &log.value)?;
             }
```
reuvenpo, line 108: I believe you don't need the `&mut` in `&mut ok.log` or the `&`s in `&log.key` and `&log.value`

cosmwasm/packages/wasmi-runtime/src/cosmwasm/types.rs
```diff
@@ -203,3 +307,18 @@ pub fn log(key: &str, value: &str) -> LogAttribute {
...
+pub enum CustomMsg {
+    Debug(String),
+    Raw(Binary),
+}
+
+impl Into<CosmosMsg<CustomMsg>> for CustomMsg {
+    fn into(self) -> CosmosMsg<CustomMsg> {
+        CosmosMsg::Custom(self)
+    }
+}
```
reuvenpo, line 324: You should group these with the definitions above

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
reuvenpo: same below

cosmwasm/packages/wasmi-runtime/src/cosmwasm/types.rs
toml01: Yep I think the current way is more readable

cosmwasm/packages/wasmi-runtime/src/cosmwasm/types.rs
toml01: Yes, keep in mind this is a copy of the original enum from `cosmwasm-std`, and should be in correlation to it.
I will add a comment describing that.

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
toml01: Maybe `(*msg).clone()` ? because `SecretMessage::from_base64` takes a `String` but `msg` is `&mut String`

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
```diff
@@ -65,99 +66,89 @@ pub fn encrypt_output(
...
+            msg_to_pass.encrypt_in_place()?;
+            *msg = b64_encode(&msg_to_pass.to_slice());
+        }
+        WasmMsg::Instantiate { msg, .. } => {
+            let mut msg_to_pass =
+                SecretMessage::from_base64((*msg).to_string(), nonce, user_public_key)?;
+
+            msg_to_pass.encrypt_in_place()?;
+            *msg = b64_encode(&msg_to_pass.to_slice());
+        }
```
reuvenpo, line 150: The section:
```rust
let mut msg_to_pass =
  SecretMessage::from_base64((*msg).to_string(), nonce, user_public_key)?;

msg_to_pass.encrypt_in_place()?;
*msg = b64_encode(&msg_to_pass.to_slice());
```
repeats twice here. you should maybe wrap it up in a function, or (and i don't know if it works) try to write the match arm as:
```
WasmMsg::Execute { msg, .. } | WasmMsg::Instantiate { msg, .. } => { ... }
```

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
toml01: `&mut ok.log` is essential because I change `log`'s value.
The other `&`s are because `encrypt_serializeable` takes `&T`

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
reuvenpo: The fact that it's `&mut String` doesn't matter. the `.` (period) operator follows references, so the `.clone()` will apply to the String behind the reference, and return a new one by value.

cosmwasm/packages/wasmi-runtime/src/cosmwasm/types.rs
reuvenpo: Mhm.
I think there already is a comment in the top of the file

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
reuvenpo: does it complain if you remove them?
`ok` here is already a `&mut T`, so so should `ok.log`, `log.key`, and `log.value` be.
And `&mut` i usable as a `&`

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
toml01:
```
WasmMsg::Execute { msg, .. } | WasmMsg::Instantiate { msg, .. } => { ... }
```
This has just changed my life, thanks :stuck_out_tongue:

assafmo:
Nice!

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
reuvenpo: awesome, haha :D

cosmwasm/packages/wasmi-runtime/src/cosmwasm/types.rs
toml01: Oh you're right it is.
I updated it and will keep the specific comments just in case :)

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
toml01: Yes it complains. It seems like the `for` statement loses the `&mut` on `ok`, maybe because I refer to the inner `ok.log`?

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
toml01: Huh, seems like it compiles just fine, but the IDE gives `mismatched types [E0308] expected `String`, found `&mut String``

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
reuvenpo: hmm... might be.
the compiler has a bunch of specific behaviours in these contexts, so i was hoping it would be smart enought to understand these things. oh well :P

cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
reuvenpo: I say ignore the IDE :D
seriously though, if the IDE gives an error over something so simple, just go with the IDE

reuvenpo: So `(*msg).clone()` is ok

PR #449, opened 2020-08-05
Gas humanize_address & canonicalize_address usage in WASM

assafmo:


cosmwasm/packages/wasmi-runtime/src/wasm/gas.rs
```diff
@@ -26,6 +26,10 @@ pub struct WasmCosts {
     pub opcodes_mul: u32,
     /// Cost of wasm opcode is calculated as TABLE_ENTRY_COST * `opcodes_mul` / `opcodes_div`
     pub opcodes_div: u32,
+    /// Cost invoking humanize_address from WASM
+    pub humanize_address: u32,
+    /// Cost invoking canonicalize_address from WASM
+    pub canonicalize_address: u32,
```
reuvenpo, line 32: I'd prefix these with `external_` or `custom_` or `cosmwasm_`

assafmo: Okay, will do

PR #471, opened 2020-08-12
Init enclave fix

Cashmaney:
Don't overwrite the original registration if `init-enclave` is ran a second time (can be done with a `--reset` flag)


cmd/secretd/attestation.go
```diff
@@ -33,18 +36,42 @@ blockchain. Writes the certificate in DER format to ~/attestation_cert
...
 		Args: cobra.ExactArgs(0),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			_, err := api.KeyGen()
-			if err != nil {
-				return fmt.Errorf("failed to initialize enclave: %w", err)
+
+			sgxSecretsPath := os.Getenv("SCRT_SGX_STORAGE")
+			if sgxSecretsPath == "" {
+				sgxSecretsPath = os.ExpandEnv("$HOME/.sgx_secrets")
+			}
```
reuvenpo, line 43: Ideally these would be in constants, but it doesn't matter right now

PR #474, opened 2020-08-13
Updated secretcli

mohammedpatla:
Need `--from` or it won't verify.


mohammedpatla:
Note on this :
In the Cosmos-Doc, they don't have the --from attribute either. So not sure fits a SDK-Doc issue or that secretcli was changed.

PR #479, opened 2020-08-13
make WasmMsg::Instantiate.label String instead of Option<String>

assafmo:


assafmo:
This is just to align this type inside the enclave with the std type which was fixed earlier this week.

assafmo:
The enclave doesn't use this field so nothing else is changed in this PR.

Issue #372, opened 2020-07-09
Secret JS CLI


levackt:
The [CLI](https://github.com/enigmampc/SecretNetwork/tree/develop/cosmwasm-js/packages/cli) doesn't work for secret contracts yet. 
You can query block height and such, but not contracts which fail with "Thrown: Error: illegal base64 data at input byte 0 (HTTP 500)"

This is expected because I'm not using the Secret JS package, so maybe there's some docs and package config needed to get this working.

reuvenpo:
done a few months ago :)

PR #481, opened 2020-08-17
Secret JS CLI #372

levackt:
By applying this change I got the [cli working](https://github.com/enigmampc/enigma-blockchain-contracts-guide/blob/master/cosmwasm-js.md)

Kicking off a draft in case anyone else is keen to get going like this already, open to any other changes you might have in mind.


assafmo:
Thanks for this. I'll reference this PR when I get to #372

PR #494, opened 2020-08-24
Parameter verification fixes

toml01:


cosmwasm/packages/wasmi-runtime/src/crypto/multisig.rs
```diff
@@ -152,3 +151,113 @@ fn decode_multisig_signature(raw_blob: &[u8]) -> Result<MultisigSignature, Crypt
...
+            0, 0, 0, 0, 0, 0, 0, 0x12, 205, 1, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
+        ];
+
+        let result = decode_multisig_signature(malformed_sig.as_slice());
+        assert!(
+            result.is_err(),
+            "Signature is: {:?} and result is: {:?}",
+            malformed_sig,
+            result
+        );
```
reuvenpo, line 222: you can probably assert equality with a specific error, if the error type implements `PartialEq` and either `Debug` or `Display` (not sure which one). you can import a quick and dirty `Display` derive macro from `derive_more::Display` which is available

reuvenpo (commented):
great fixes so far :muscle::smile:
Just left some minor aesthetic comments on the tests.

cosmwasm/packages/wasmi-runtime/src/crypto/multisig.rs
```diff
@@ -152,3 +151,113 @@ fn decode_multisig_signature(raw_blob: &[u8]) -> Result<MultisigSignature, Crypt
 
     Ok(signatures)
 }
+
+#[cfg(feature = "test")]
+pub mod tests_decode_multisig_signature {
+    use crate::crypto::multisig::{decode_multisig_signature, MultisigSignature};
+
```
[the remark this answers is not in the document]
toml01: Yep, I renamed it so I could create another test module in the future, e.g. `pub mod tests_multisig_pubkey`

cosmwasm/packages/wasmi-runtime/src/crypto/multisig.rs
toml01: Not sure that it makes a difference lol

toml01: besides, all errors there are `CryptoError::ParsingError` so..

reuvenpo: The main difference is in the error message if it fails, but yeah this isn't really important in this case.

cosmwasm/packages/wasmi-runtime/src/crypto/multisig.rs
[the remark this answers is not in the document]
reuvenpo: fair enough i suppose.

x/compute/internal/keeper/keeper.go
```diff
@@ -124,37 +124,56 @@ func GetSignerSignature(signer exported.Account, tx auth.StdTx) (authtypes.StdSi
...
+	// https://github.com/cosmos/cosmos-sdk/commit/c13809062ab16bf193ad3919c77ec03c79b76cc8#diff-a64b9f4b7565560002e3ac4a5eac008bR148
+	tx := authtypes.StdTx{}
+	txBytes := ctx.TxBytes()
+	err := k.cdc.UnmarshalBinaryLengthPrefixed(txBytes, &tx)
+	if err != nil {
+		return authtypes.StdSignature{
+			PubKey:    secp256k1.PubKeySecp256k1{},
+			Signature: []byte{},
+		}, nil, sdkerrors.Wrap(types.ErrInstantiateFailed, fmt.Sprintf("Unable to decode transaction from bytes: %s", err.Error()))
+	}
```
reuvenpo, line 138: To make these slightly nicer, you can assign the empty `authtypes.StdSignature` to a local in the beginning of the function, and just refer to it in the returns.

reuvenpo (approved):
:100:

Issue #485, opened 2020-08-22
Add secret-20 support in the CLI


Cashmaney:
Secret-20 should be natively supported in the CLI. It'll go a long way to making it both more useable and looking more legit

Cashmaney:
> Do we want to also include it in SecretJS to make web integration easy?

Maybe, but it's a different use case. SecretJS is going to be used by developers, who don't need as much hand-holding vs the CLI. For example, web3 doesn't support "native" erc-20 functions. You just import the contract ABI and call functions

PR #495, opened 2020-08-25
Secret-secret cli

Cashmaney:
Adding some secret token commands to the cli #485


cmd/secretcli/secret20.go
```diff
@@ -0,0 +1,371 @@
...
+	return s20TxCmd
+}
+
+func S20TransferHistoryCmd(cdc *codec.Codec) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "history [contract address] [account] [viewing_key]",
+		Short: "View your transaction history",
+		Long:  `Print out all transactions you have been a part of - either as a sender or recipient`,
+		Args:  cobra.ExactArgs(3),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
reuvenpo, line 69: I'd add a `-n` flag for specifying the maximum amount of txes to display

cmd/secretcli/secret20.go
```diff
@@ -0,0 +1,371 @@
...
+			toAddr, err := sdk.AccAddressFromBech32(args[1])
+			if err != nil {
+				return errors.New("invalid recipient address")
+			}
+
+			amount := args[2]
+			_, err = strconv.ParseUint(amount, 10, 64)
+			if err == nil {
+				return errors.New("invalid amount format")
+			}
```
reuvenpo, line 192: Does Secret20 only work with whole numbers as amounts?

cmd/secretcli/secret20.go
```diff
@@ -0,0 +1,371 @@
...
+	return []byte(fmt.Sprintf("{\"withdraw\": {\"amount\": \"%s\"}}", amount))
+}
+
+func createViewingKeyMsg(data string) []byte {
+	return []byte(fmt.Sprintf("{\"create_viewing_key\": {\"entropy\": \"%s\"}}", data))
+}
+
+func setViewingKeyMsg(data string) []byte {
+	return []byte(fmt.Sprintf("{\"set_viewing_key\": {\"key\": \"%s\"}}", data))
+}
```
reuvenpo, line 371: It's probably a good idea to use proper json escaping here, using a json library, but bad inputs will just cause the contract to fail, so it's probably ok?

cmd/secretcli/secret20.go
```diff
@@ -0,0 +1,371 @@
...
+func s20SetViewingKey(cdc *codec.Codec) *cobra.Command {
+
+	cmd := &cobra.Command{
+		Use:   "set-viewing-key [contract address or label]",
+		Short: "sets the viewing key for your account",
+		Long: `This command is useful if you want to manage multiple secret tokens with the same viewing key. WARNING:
+This should only be used to duplicate keys created with the create-viewing-key command, or if you really really know what
+you're doing`,
+		Args: cobra.MinimumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
reuvenpo, line 255: I'd add a newline before the warning, to make it easier to notice in the `--help` details

reuvenpo: also, the `Use` field doesn't mention the viewing key in the second argument

cmd/secretcli/secret20.go
```diff
@@ -0,0 +1,371 @@
...
+			msg := depositMsg()
+
+			amountStr := viper.GetString(flagAmount)
+
+			return cli.ExecuteWithData(cmd, contractAddr, msg, amountStr, false, "", "", cdc)
+		},
+	}
+	cmd.Flags().String(flagAmount, "", "")
+	return cmd
+}
```
reuvenpo, line 310: why did you decide to pass the amount to `deposit` as a flag, rather than as a positional argument, like in `withdraw`?

x/compute/client/cli/query.go
```diff
@@ -438,79 +440,108 @@ func GetCmdQuery(cdc *codec.Codec) *cobra.Command {
 		Use:   "query [bech32_address] [query]", // TODO add --from wallet
 		Short: "Calls contract with given address  with query data and prints the returned result",
 		Long:  "Calls contract with given address  with query data and prints the returned result",
-		Args:  cobra.ExactArgs(2),
-		RunE: func(_ *cobra.Command, args []string) error {
-			cliCtx := context.NewCLIContext().WithCodec(cdc)
+		Args:  cobra.MinimumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			inBuf := bufio.NewReader(cmd.InOrStdin())
+			cliCtx := context.NewCLIContextWithInput(inBuf).WithCodec(cdc)
```
reuvenpo, line 446: We should document this extension in the `Use`, `Short`, or `Long` field.

x/compute/client/cli/query.go
```diff
@@ -438,79 +440,108 @@ func GetCmdQuery(cdc *codec.Codec) *cobra.Command {
...
+	cmd.Flags().String(flagLabel, "", "A human-readable name for this contract in lists")
+	return cmd
+}
 
-			codeHash, err := GetCodeHashByContractAddr(cliCtx, addr)
-			if err != nil {
-				return fmt.Errorf("contract not found: %s", addr)
-			}
+func QueryWithData(_ *cobra.Command, contractAddress string, cdc *codec.Codec, queryData []byte) error {
+	cliCtx := context.NewCLIContext().WithCodec(cdc)
```
reuvenpo, line 486: why require `_ *cobra.Command`?

reuvenpo (commented):
maybe we should also provide aliases for viewing keys like there already is for account keys? Is that feasible? because the current design means people have to put their viewing keys in environment variables if they don't want them to be saved to terminal hostory/

cmd/secretcli/secret20.go
Cashmaney: Yes

cmd/secretcli/secret20.go
Cashmaney: The checks you added that it is a valid json will probably make this fail gracefully anyway

cmd/secretcli/secret20.go
Cashmaney: because you actually need to send native coins

reuvenpo: Ah, so we want the commands to look more like the native commands?

cmd/secretcli/secret20.go
reuvenpo: That's fair.
I still find it ugly, but it works so I'm not gonna pursue this right now.

cmd/secretcli/secret20.go
reuvenpo: cool

cmd/secretcli/secret20.go
Cashmaney: A deposit is literally "send SCRT and get sSCRT", so you have to have it actually work:)

toml01 (approved):

PR #496, opened 2020-08-25
updating dev docs

jlwaugh:


docs/dev/developing-secret-contracts.md
```diff
@@ -0,0 +1,157 @@
...
+Example Invocation from `secretcli`:
+
+```bash
+secretcli tx compute instantiate "$CODE_ID" "$INPUT_MSG" --label "$UNIQUE_LABEL" --from "$MY_KEY"
+```
+
+Example Invocation from `SecretJS`:
+
+```js
+// TODO
```
levackt, line 53: Before you can init here you'll need funded account and a client connection, I don't know if that's worth a mention.
All the snippets in the TODOs can come from [here](https://github.com/enigmampc/enigma-blockchain-contracts-guide/blob/testnet-quickstart/cosmwasm-js.md)

I'd suggest you try it too, but currently the cli won't work unless you [install it locally from here](https://github.com/enigmampc/SecretNetwork/pull/481/files). That said the code and config works as it is outside the cli, eg holdem, voting app etc

assafmo: You're right. This is just supposed to be a context-less code snippet, to show to devs that it's really easy to do.

docs/protocol/governance.md
```diff
@@ -115,12 +115,12 @@ You can see another `param-change` example here: [enigma-1-proposal-3.json](<[/d
...
+| `mint`         | `InflationMax`            | string (dec)     | `"0.150000000000000000"`                                                                                  |
 | `mint`         | `InflationMin`            | string (dec)     | `"0.070000000000000000"`                                                                                  |
 | `mint`         | `GoalBonded`              | string (dec)     | `"0.670000000000000000"`                                                                                  |
 | `mint`         | `BlocksPerYear`           | string (uint64)  | `"6311520"`                                                                                               |
-| `slashing`     | `SignedBlocksWindow`      | string (int64)   | `"100"`                                                                                                   |
+| `slashing`     | `SignedBlocksWindow`      | string (int64)   | `"5000"`                                                                                                   |
```
assafmo, line 123: These are just examples, no need to keep them updated with mainnet.

assafmo (approved):

Issue #133, opened 2020-04-05
Vulnerability in ocall_read_db


reuvenpo:
While reviewing some things around the enclave boundary i realized there is a data-leakage vulnerability in `ocall_read_db`.
Since the enclave is run by untrusted hosts, it can not trust values returned from that host. This means inputs to ecalls, can return values of ocalls.
The `ocall_read_db` method is part of the following sequence of operations:
* The host calls the `ecall_init`/`ecall_handle`/`ecall_query` function
* A WASM runtime is started with a module and the appropriate function in it is called
* The WASM module requests to read a from the KV storage (DB)
* The `ocall_read_db` function is called
* The KV storage fetches the value of the key and returns it
* The host calls `ecall_allocate` to copy the value into the enclave memory
* `ecall_allocate` allocates a `Box<Vec<u8>>` and returns it as `EnclavePointer`
* This pointer is then returned as is from `ocall_read_db` and the enclave retrieves the vector.
* The buffer is copied into the WASM memory space where the secret contract can read form it.

Now, if a malicious host instead returns a pointer to something that _looks_ like a vector to an area of memory inside the enclave that should be kept secret, the contract will read it instead, treat it as the return value of the `read_db` operation, and might include it in the JSON it returns out of the enclave when it finishes running (and `ecall_init`/`ecall_handle`/`ecall_query` return from the enclave).
If done accurately, this may leak either sensitive user data, secret keys, or anything else that can gradually break the whole security network.

There are two solutions to this:
1. We encrypt the pointer that returns from `ecall_allocate` - but then which keys should we use for that?
2. Add a `hash` field to `EnclaveBuffer` that shows the hash of the data inside the buffer. That way, an attacker will have to know ahead of time what information he expects to find behind the pointer (which isn't great) and can still cause Use-After-Free or other allocator errors errors if he gives us dubious addresses.

Ideally, we would employ both techniques in tandem.

reuvenpo:
Another option is to have a static global variable with a `HashSet<*mut c_void>` (a set of pointers) which would keep track of which addresses represent memory that was allocated but not yet freed. This way if `ocall_read_db` returns an address that is not recognized, we try to call it again. This way the worst case scenario is that the host forces us into an infinite loop that allocates more and more memory, until the enclave crashes from an allocator fault.

reuvenpo:
[...]

PR #501, opened 2020-08-26
added allocation stack to help validate EnclaveBuffers returned from …

reuvenpo:
This should close the last problems presented in #133. 
One question remains - if the pointer is not recognized, should we return `None` (current implementation, represents a valid, empty response), Or should we return some kind of `EnclaveError`, and fail, because something is wrong or the host is trying to mess with us?
CC: @assafmo


reuvenpo:
commit 9c43b139 shows how we can report an error if the host is caught cheating.

assafmo:
> One question remains - if the pointer is not recognized, should we return `None` (current implementation, represents a valid, empty response), Or should we return some kind of `EnclaveError`, and fail, because something is wrong or the host is trying to mess with us?

`EnclaveError` IMO.

assafmo (approved):

PR #507, opened 2020-08-27
Remove wasmd's access config

assafmo:


Cashmaney (approved):

x/compute/internal/keeper/keeper.go
```diff
@@ -125,14 +125,15 @@ func (k Keeper) setParams(ctx sdk.Context, ps types.Params) {
...
-		}
+			   	return k.create(ctx, creator, wasmCode, source, builder, &types.AccessConfig{Type: types.Everybody}  , k.authZPolicy )
+			   }
+
+			   func (k Keeper) create(ctx sdk.Context, creator sdk.AccAddress, wasmCode []byte, source string, builder string, instantiateAccess *types.AccessConfig ) (codeID uint64, err error) {
+		/*
+			   		if !authZ.CanCreateCode(k.getUploadAccessConfig(ctx), creator) {
+			   			return 0, sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "can not create code")
+			   		}
 	*/
```
reuvenpo, line 137: Adding so much whitespace is only going to make merges harder :scream: 
You want to just add a `/*` on its own line before the content, and a `*/` on its own line after the content...

x/compute/internal/types/genesis.go
```diff
@@ -20,16 +20,16 @@ func (s Sequence) ValidateBasic() error {
...
 	Sequences []Sequence `json:"sequences,omitempty"`
 }
 
 func (s GenesisState) ValidateBasic() error {
-	if err := s.Params.ValidateBasic(); err != nil {
+	/* 	if err := s.Params.ValidateBasic(); err != nil {
 		return sdkerrors.Wrap(err, "params")
-	}
+	} */
 	for i := range s.Codes {
```
reuvenpo, line 33: similarly here and other places in the PR.
Git only gets it if you put it on separate lines...

reuvenpo (changes requested):

x/compute/internal/keeper/keeper.go
Cashmaney: or just delete this stuff since we don't care what they change in wasmd 😛

PR #513, opened 2020-08-31
Azure touchups

assafmo:


reuvenpo:
are we basically dropping support for the azure templates for `enigma-testnet-1`?

Cashmaney:
> are we basically dropping support for the azure templates for `enigma-testnet-1`?

`enigma-testnet-1` is our internal testnet from a couple of months ago..

reuvenpo:
Oh, sorry, i got the chain ids mixed up in my mind

PR #517, opened 2020-09-01
Added recursion limit of 5 to queries

### LINKED ISSUE #504: Fix OOM handler once and for all

reuvenpo:
This should mitigate OOM cases by putting a hard cap on the depth of recursive queries allowed for secret contracts.
**This PR sets the enclave memory to 128MiB in both test and prouction modes, and then sets the recursion limit to 5.**
Thisis because we estimate that the maximum memory consumption per level of recursion in the enclave is approximately 20MiB, and then we add some overhead in case memory requirements are slightly higher for any reason.
This PR hopefully closes #504 

TO REVIEWERS: please note the change to `cosmwasm_std::SystemError`

UPDATE: Enclave now has 256MiB heap, but we keep the recursion limit at 5.


cosmwasm/packages/wasmi-runtime/src/recursion_depth.rs
```diff
@@ -0,0 +1,34 @@
+use std::sync::SgxMutex;
+
+use lazy_static::lazy_static;
+
+use enclave_ffi_types::EnclaveError;
+
+const RECURSION_LIMIT: i32 = 5;
```
assafmo, line 7: Why not `u32`?

cosmwasm/packages/wasmi-runtime/src/recursion_depth.rs
```diff
@@ -0,0 +1,34 @@
...
+        return Err(EnclaveError::ExceededRecursionLimit);
+    }
+    *depth += 1;
+    Ok(())
+}
+
+pub fn decrement() {
+    let mut depth = RECURSION_DEPTH.lock().unwrap();
+    *depth -= 1;
+}
```
assafmo, line 29: I think this effectively allows 6 VMs in parallel, right?
I'd feel more comfortable if we only allow 5 VMs in parallel and keep the memory at 128 MiB.

cc @toml01 @Cashmaney

assafmo: I.e set `RECURSION_LIMIT = 4`.

x/compute/internal/keeper/secret_contracts_test.go
```diff
@@ -1287,6 +1288,7 @@ func TestMsgSenderInCallback(t *testing.T) {
 }
 
 func TestInfiniteQueryLoopKilledGracefullyByOOM(t *testing.T) {
+	t.SkipNow() // We no longer expect to hit OOM trivially
```
assafmo, line 1291: :+1:

cosmwasm/packages/wasmi-runtime/Enclave.config.xml
```diff
@@ -3,7 +3,7 @@
   <ProdID>0</ProdID>
   <ISVSVN>0</ISVSVN>
   <StackMaxSize>0x800000</StackMaxSize>
-  <HeapMaxSize>0x4000000</HeapMaxSize>
+  <HeapMaxSize>0x8000000</HeapMaxSize>
```
Cashmaney, line 6: Can we increase to 256 to be extra safe?

cosmwasm/packages/wasmi-runtime/src/exports.rs
```diff
@@ -116,6 +116,10 @@ pub unsafe extern "C" fn ecall_init(
     sig_info: *const u8,
     sig_info_len: usize,
 ) -> InitResult {
+    if let Err(err) = recursion_depth::increment() {
+        error!("recursion limit exceeded, can not perform query!");
```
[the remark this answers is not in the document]
Cashmaney: You could make a contract that inits itself during the init, so I guess so.

cosmwasm/packages/wasmi-runtime/src/exports.rs
```diff
@@ -200,6 +205,10 @@ pub unsafe extern "C" fn ecall_handle(
     sig_info: *const u8,
     sig_info_len: usize,
 ) -> HandleResult {
+    if let Err(err) = recursion_depth::increment() {
+        error!("recursion limit exceeded, can not perform query!");
```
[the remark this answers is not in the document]
Cashmaney: Would probably be more clear to change the name of `query` to `handle`

Also, can we make these `warn` instead of `error`? This is expected behaviour, so not really worth an error

cosmwasm/packages/enclave-ffi-types/src/types.rs
```diff
@@ -154,6 +154,8 @@ pub enum EnclaveError {
     Panic,
     #[display(fmt = "enclave ran out of heap memory")]
     OutOfMemory,
+    #[display(fmt = "query recursion limit exceeded")]
```
Cashmaney, line 157: This is what is going to be returned to the contract caller, right? Can we make this `depth of nested contract calls exceeded`?

cosmwasm/packages/wasmi-runtime/src/exports.rs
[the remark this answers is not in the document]
reuvenpo: @Cashmaney initializing a new contract will only happen after the first init completes, and that doesn't count towards recursion.
@assafmo Not sure what you mean by your question, but I believe that this error condition is currently unreachable. I think we can safely remove it completely right now, and have `recursion_depth::increment()` simply increment the counter with no further checks, but i wanted to stay on the safe side here, in case something changes in the future, and we can easily spot that we forgot to add a limit somewhere.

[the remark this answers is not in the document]
reuvenpo: we should rephrase the error message as `"can not perform init"` ,as @Cashmaney suggests

cosmwasm/packages/wasmi-runtime/src/exports.rs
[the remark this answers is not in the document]
reuvenpo: I agree with you @Cashmaney

cosmwasm/packages/wasmi-runtime/src/recursion_depth.rs
reuvenpo: No reason, really. It's the default integer type. our limit is so low that this can be any integer type

cosmwasm/packages/wasmi-runtime/src/recursion_depth.rs
reuvenpo: No. This counter also includes the initial call into the enclave, so a recursion limit of 2 means you can call `query_chain` once. As you've seen, I increment the counter in the `init` and `handle` entry points as well.
It effectively tracks how many running instances of wasmi we have.

cosmwasm/packages/wasmi-runtime/Enclave.config.xml
reuvenpo: sure. Do you want to adjust the recursion limit to 10 as well, or do you want to have a huge margin of error?

cosmwasm/packages/enclave-ffi-types/src/types.rs
reuvenpo: yeah, sure. it does sound better :) 
keep in mind though that I think this error will never be displayed.

cosmwasm/packages/wasmi-runtime/src/exports.rs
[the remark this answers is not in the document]
reuvenpo: actually... This could even be `debug` since node runners probably won't care about this at all.
On the other hand this isn't exactly hidden because the host can theoretically track how many recursions are done per tx

cosmwasm/packages/wasmi-runtime/Enclave.config.xml
Cashmaney: Huge margin of error:)

cosmwasm/packages/enclave-ffi-types/src/types.rs
Cashmaney: I'm thinking of dev environments where you can make mistakes - I had that happen to me a few times

cosmwasm/packages/wasmi-runtime/Enclave.config.xml
reuvenpo: cool :)

cosmwasm/packages/wasmi-runtime/src/exports.rs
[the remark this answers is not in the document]
assafmo: Yep. I'd add a code comment with what you just wrote:
> I believe that this error condition is currently unreachable. I think we can safely remove it completely right now, and have `recursion_depth::increment()` simply increment the counter with no further checks, but i wanted to stay on the safe side here, in case something changes in the future, and we can easily spot that we forgot to add a limit somewhere.

cosmwasm/packages/wasmi-runtime/src/exports.rs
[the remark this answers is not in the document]
assafmo: This isn't expected behavior, `init` and `handle` are always on `depth=0`.
See [my other comment](https://github.com/enigmampc/SecretNetwork/pull/517#discussion_r481971747) about adding a code comment here as well.

cosmwasm/packages/wasmi-runtime/src/recursion_depth.rs
assafmo: Sorry I meant to comment this about `RECURSION_DEPTH`.
I'm trying to think whether an attacker can screw with `RECURSION_DEPTH` and somehow get to a negative count.

I'd turn them both to `u8` and use `saturating_add` and `saturating_sub`.

Also, let's assume that somehow queries and txs work concurrently, E.g. handle called a sub-query, and while inside the `ocall` the program decides to switch to a query thread. Now two different contexts share the same counter.
The counter protects against 5 (6?) VMs in parallel, but now the tx might be cut short because of a local query and this will kick the node out of consensus.
Do we want to protect against that as well, or are we good with this?

cosmwasm/packages/wasmi-runtime/src/recursion_depth.rs
assafmo: Okay so let's say I'm calling and endless loop of sub query.

- 1st call: 0 VMs total => `depth (0) != RECURSION_LIMIT (5)` => `depth <- 1`
- 2nd call: 1 VMs total => `depth (1) != RECURSION_LIMIT (5)` => `depth <- 2`
- 3th call: 2 VMs total => `depth (2) != RECURSION_LIMIT (5)` => `depth <- 3`
- 4th call: 3 VMs total => `depth (3) != RECURSION_LIMIT (5)` => `depth <- 4`
- 5th call: 4 VMs total => `depth (4) != RECURSION_LIMIT (5)` => `depth <- 5`
- 6th call: 5 VMs total => `depth (5) == RECURSION_LIMIT (5)` => `error`

Cool, you're right.

cosmwasm/packages/wasmi-runtime/Enclave.config.xml
assafmo: @Cashmaney will 256 MiB work in the CIs?

cosmwasm/packages/wasmi-runtime/src/exports.rs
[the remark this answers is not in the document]
assafmo: @Cashmaney the goal here is to prevent more than 5 VMs allocated at the same time.

The only 3 possible combinations are:

- init + query + query + query + query
- handle + query + query + query + query
- query + query + query + query + query

cosmwasm/packages/wasmi-runtime/src/recursion_depth.rs
reuvenpo: I would like to protect against this, but that would require thread local variables (which we should definitely try to tackle again).
In the meantime though, the enclave only has one TCS so only one thread can interact with it at a time (I think?)

assafmo (approved):
Excellent work @reuvenpo!

PR #523, opened 2020-09-02
Additional logs safety & tests

toml01:


Cashmaney (approved):

cosmwasm/packages/wasmi-runtime/Enclave.edl
```diff
@@ -10,6 +10,7 @@ enclave {
     from "sgx_tprotected_fs.edl" import *;
 
     from "sgx_env.edl" import u_getenv_ocall;
+    from "sgx_env.edl" import u_setenv_ocall;
```
reuvenpo, line 14: Can this be on one line?

cosmwasm/packages/wasmi-runtime/src/lib.rs
```diff
@@ -67,6 +63,13 @@ fn log_level_from_str(env_log_level: &str) -> Option<LevelFilter> {
...
 
 fn set_log_level_or_default(default: LevelFilter, max_level: LevelFilter) {
+    if default > max_level {
+        panic!(
+            "Logging configuration is broken, stopping to prevent secret leaking. default: {:?}, max level: {:?}",
+            default, max_level
+        );
+    }
+
     let mut log_level = default;
```
reuvenpo, line 73: :100:

cosmwasm/packages/wasmi-runtime/src/lib.rs
```diff
@@ -78,7 +81,50 @@ fn set_log_level_or_default(default: LevelFilter, max_level: LevelFilter) {
...
+            panic!("{}: {} tests failed", file!(), failures);
+        }
+    }
+
+    fn test_log_level() {
+        env::set_var("LOG_LEVEL", "WARN");
+        set_log_level_or_default(LevelFilter::Error, LevelFilter::Info);
+        assert_eq!(log::max_level(), LevelFilter::Warn);
+        info!("If you see this, logging is not working correctly!"); // This is not ideal, but checking stdout will be an overkill
+
```
reuvenpo, line 112: Is it easy to maybe implment your own logger that collects log messages to a `Vec<String>`?
You can then check its contents

assafmo (approved):

cosmwasm/packages/wasmi-runtime/src/lib.rs
toml01: That's a good idea, but the point is to check **our** implementation of the logger. Having a custom logger impl. for tests is not ideal imo

reuvenpo: Yeah, but this test check the behaviour of the logging system and whether or not it routes messages correctly, not the specific logger, so for these specific purposes, I think it's ok.

toml01: Yeah thats a fair point

PR #531, opened 2020-09-05
Switched TCSPolicy to bound mode, to ensure thread affinity

reuvenpo:
That is, this makes sure that when a specific thread communicates with the enclave, another thread cannot hijack the session with the enclave and try to mess with it in some way. This should prevent some exploits.

This commit also changes some global variables to thread_local to further protect data leakage between threads.

Additionally, TCSNum is increased to 8. This is mostly to accomodate the Go test runner, but also to enable queries to be performed on running nodes.

Please note that we keep the recursion limit counter as a usual static mutex, since the amount of allowed VMs is a global restriction, not a per-thread restriction.

EDIT: The branch name is unrelated - I mixed up my branch names. oh well :man_shrugging:


cosmwasm/packages/wasmi-runtime/Makefile
```diff
@@ -108,8 +108,8 @@ ffi-types:
 	RUST_TARGET_PATH=$(Rust_Target_Path) RUSTFLAGS=$(Rust_Flags) xargo build --features "build_headers" --target x86_64-unknown-linux-sgx -Z unstable-options --profile $(BUILD_PROFILE) && \
 	cp $(Rust_Target_Path)/../../../target/headers/enclave-ffi-types.h $(Rust_Target_Path)/../target/headers/
 
-librust_cosmwasm_enclave.signed.so: librust_cosmwasm_enclave.so
-	sgx_sign sign -key Enclave_private.pem -enclave $^ -out $@ -config $(ENCLAVE_CONFIG)
+librust_cosmwasm_enclave.signed.so: librust_cosmwasm_enclave.so $(ENCLAVE_CONFIG)
+	sgx_sign sign -key Enclave_private.pem -enclave $< -out $@ -config $(ENCLAVE_CONFIG)
```
reuvenpo, line 113: Note that whether or not we accept this PR, we should include this specific change.

assafmo: So please extract it to a new micro-PR.

reuvenpo: https://github.com/enigmampc/SecretNetwork/pull/537

reuvenpo:
This latest commit helps somewhat, but under high loads i still get OUT_OF_TCS for some reason. 🤔

Cashmaney:
Closing in favor of #879

PR #539, opened 2020-09-09
Use sent_funds in callback_sig

toml01:


docs/dev/privacy-model-of-secret-contracts.md
```diff
@@ -39,6 +39,29 @@ Secret Contract developers must always consider the trade-off between privacy, u
...
+As an example, let's say we are implementing an admin interface for a contract, i.e. functionality that is open only for a predefined address.
+In that case, we want to know that the `env.message.sender` parameter that is given during contract execution is legit, then we want to check that `env.message.sender == predefined_address` and provide admin functionality if that condition is met.
+If the `env.message.sender` parameter can be tampered with - we effectively can't rely on it and cannot implement the admin interface.
+
+## Tx Parameter Verification
+Some parameters are easier to verify, but for others it is less trivial to do so. Exact details about individual parameters are detailed further in this document.
+
+The parameter verification method depends on the contract caller:
+ - If the contract is called by a transaction (i.e. someone sends a compute tx) we use the already-signed transaction and verify it's data inside the enclave. More specifically:
+   - Verify signature inside the enclave.
```
reuvenpo, line 57: I'd rephrase this point as"Verify that the transaction was signed by its sender"

docs/dev/privacy-model-of-secret-contracts.md
```diff
@@ -39,6 +39,29 @@ Secret Contract developers must always consider the trade-off between privacy, u
...
+   - Verify that the parameters sent to the enclave matches with the signed data.
+ - If the contract is called by another contract (i.e. we don't have a signed tx to rely on) we create a callback signature, effectively signing the parameters sent to the next contract:
+   - Caller contract creates `callback_signature` based on parameters it sends, passes it on to the next contract.
+   - Receiver contract creates `callback_signature` based on the parameter it got.
+   - Receiver contract verifies that the signature it created matches the signature it got from the caller.
+   - For the specifics, visit the [encryption specs](../protocol/encryption-specs.md#Output).
+
```
reuvenpo, line 64: I'd mention that this is a signature that can only be created inside the enclave, since it uses a secret that is kept inside the enclace.

docs/dev/privacy-model-of-secret-contracts.md
```diff
@@ -39,6 +39,29 @@ Secret Contract developers must always consider the trade-off between privacy, u
...
+During execution, some contracts may want to use "external-data" - meaning data that is generated outside of the enclave and sent into the enclave - such as the tx sender address, the funds sent with the tx, block height, etc..
+As these parameters get sent to the enclave, they can theoretically be tampered with, and an attacker might send false data.
+Thus, relying on such data might be risky.
+
+As an example, let's say we are implementing an admin interface for a contract, i.e. functionality that is open only for a predefined address.
+In that case, we want to know that the `env.message.sender` parameter that is given during contract execution is legit, then we want to check that `env.message.sender == predefined_address` and provide admin functionality if that condition is met.
+If the `env.message.sender` parameter can be tampered with - we effectively can't rely on it and cannot implement the admin interface.
+
+## Tx Parameter Verification
+Some parameters are easier to verify, but for others it is less trivial to do so. Exact details about individual parameters are detailed further in this document.
```
reuvenpo, line 53: This section is a very good high-level explanation of the limitations that contract authors need to be aware of, but it's missing the punchline of "this is what we know how to absolutely verify, but that thing we don't"

reuvenpo (changes requested):
You can avoid the `.clone()`s of the sent fund by changing `create_callback_signature` and `verify_callback_sig` to accept `&[Coin]`, and then just pass the vector of coins with a `&`, which would dereference into the slice type. You can also then remove the `&` from the call to `serde_json::to_vec()`.

docs/dev/privacy-model-of-secret-contracts.md
toml01: This is the purpose of the following table:
https://github.com/enigmampc/SecretNetwork/blob/2aa31ad892fba390226665b3d29d021b0f6c6143/docs/dev/privacy-model-of-secret-contracts.md#inputs

reuvenpo: Ahhh, true true, fair enough

toml01:
Changes to docs were pushed directly to master @reuvenpo

reuvenpo (approved):

PR #542, opened 2020-09-10
Remove wasm from git

toml01:


toml01:
I thought it will make CI slower too, but you can check the times on this PR in comparison with others - it doesn't seems like it adds much overhead
Also, less conflicts

reuvenpo (approved):
Falcon, you're Go for launch!

assafmo (approved):

PR #545, opened 2020-09-14
Update join-validator-mainnet.md

moonstash:
There is nothing telling users how to set gas or where to set it and all transactions will fail unless this is set right.


toml01:
Thanks @moonstash !

PR #564, opened 2020-09-21
make cosmwasm-storage's dependency on cosmwasm-std more flexible

reuvenpo:
I think that our fork of `cosmwasm-storage` at `v1.0.0` is broken... trying to see if this fixes it.


reuvenpo:
Nope, problem was elsewhere

PR #578, opened 2020-09-30
Docker build fixes

reuvenpo:
The first commit here changes a use of `git branch --show-current` with `git rev-parse --abbrev-ref HEAD` because the former is [not supported on git versions before 2.22](https://stackoverflow.com/a/6245587) ([[release notes]](https://github.com/git/git/blob/master/Documentation/RelNotes/2.22.0.txt#L32)), while [Ubuntu 18.04 LTS still offers 2.17](https://packages.ubuntu.com/bionic/git).

The second commit here adds keys to `base.Dockerfile` which were missing when building for SW mode.


toml01 (approved):
:ok_hand:

PR #583, opened 2020-10-16
fixes #581 (double escaping logs)

### LINKED ISSUE #581: Unnecessary escaping in logs returned from secret contracts
@baedrik noticed an issue when trying to pass characters that are usually escaped in strings, through secret contract logs.

The flow here is:
* Pass a string to the `cosmwasm_std::log` function
* Return it from the handle/query/init function
* Read it with `secretcli q compute tx`

He observed that e.g. a literal `"` turns into a literal `\\\"`, and `\n` (a newline) turned into a literal `\\n`

This is probably a compound of two different serialization bugs in different parts of the system, because of the double escaping.

EDIT: marked as "good first issue" since it's probably a simple programming error, and could get someone new familiar with the codebase, if they want to start contributing.

baedrik:
Fixes the extra escaping that was being done with log Strings.  Also fixes how any double quotes ending a log String were getting stripped, even if the contract dev wanted their log String to be quoted

### DIFF
--- a/cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
+++ b/cosmwasm/packages/wasmi-runtime/src/wasm/io.rs
@@ -49,6 +49,23 @@ where
     Ok(b64_encode(encrypted_data.as_slice()))
 }
 
+// use this to encrypt a String that has already been serialized.  When that is the case, if
+// encrypt_serializable is called instead, it will get double serialized, and any escaped
+// characters will be double escaped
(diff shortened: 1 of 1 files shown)


Cashmaney (approved):

reuvenpo (approved):
great PR @baedrik, thanks again for putting in the effort to track down this bug!

baedrik:
You're welcome!  Happy to help out!

PR #602, opened 2020-10-20
Make the gzip in contract builds reproducible

reuvenpo:
contract builds are already reproducible, but this makes the gzip reproducible too


toml01:
So small and yet so big

toml01 (approved):

PR #624, opened 2020-11-02
remove the stale testnet page and cleanup intro

levackt:
@jlwaugh Updated the devs intro doc to remove some dated info.

We also need to sync https://build.scrt.network/dev/developers.html


toml01 (approved):
Looks good to me. Thanks @levackt

PR #630, opened 2020-11-03
Added optional signer interface

Cashmaney:


assafmo (approved):

cosmwasm-js/packages/sdk/src/wallet.ts
```diff
@@ -0,0 +1,24 @@
...
+export interface OfflineSigner {
+  /**
+   * Get AccountData array from wallet. Rejects if not enabled.
+   */
+  readonly getAccounts: () => Promise<readonly AccountData[]>;
+
+  /**
+   * Request signature from whichever key corresponds to provided bech32-encoded address. Rejects if not enabled.
+   */
+  readonly sign: (address: string, message: Uint8Array, prehashType?: PrehashType) => Promise<StdSignature>;
```
Thunnini, line 23: Can you change the sign method interface as [readonly sign: (signerAddress: string, signDoc: StdSignDoc) => Promise<SignResponse>](https://github.com/cosmos/cosmjs/blob/v0.23.1/packages/launchpad/src/signer.ts#L37) as this is more recent version that Keplr supports.

Thunnini (commented):
Sorry for the late response.

The offline signer interface of this commit is not a form of the supported offline signer in Keplr extension.
And, it seems that there is still no way to set the `enigmaUtils` manually. In the Keplr example, it set the `enigmaUtils` forcibly by ignoring the access modifier of the `restClient` https://github.com/chainapsis/keplr-example/blob/secret-wasm/src/main.js#L109. However, this way is not very proper, so, I want to suggest that the constructor of `SigningCosmWasmClient` accepts the `enigmaUtils` and set it if it is not undefined such as
```typescript
public constructor(
    apiUrl: string,
    senderAddress: string,
    signer: SigningCallback | OfflineSigner,
    enigmautils?: SecretUtils,
    seed?: Uint8Array,
    customFees?: Partial<FeeTable>,
    broadcastMode = BroadcastMode.Block,
  ) {
    super(apiUrl, seed, broadcastMode);
    this.anyValidAddress = senderAddress;
    this.senderAddress = senderAddress;
    //this.signCallback = signCallback ? signCallback : undefined;
    this.signer = signer;
    if (enigmaUtils) {
      this.restClient.enigmautils = enigmautils;
    }
    this.fees = { ...defaultFees, ...(customFees || {}) };
  }
```

cosmwasm-js/packages/sdk/src/wallet.ts
assafmo: Sorry, I totally missed this. I'll work to address your suggestions.

BTW This PR works very good for us in [this code segment](https://github.com/enigmampc/EthereumBridgeFrontend/blob/7c996239160863d278830e0ae3059020cb9c55ba/src/stores/UserStore.ts#L125-L136) with Keplr's `OfflineSigner`. Is this Okay?

assafmo: Done. I made the `SigningCosmWasmClient` constructor a bit different from what you suggested, because if you pass `enigmautils` then `seed` is irrelevant, so now you can pass one or the other, but not both.

Look at the difference from your suggestion and let me know what you think ([Link to code](https://github.com/enigmampc/SecretNetwork/blob/3f4a7d6038303200c1cc6c39f412d7e2f4bddb43/cosmwasm-js/packages/sdk/src/signingcosmwasmclient.ts#L123-L145)):
```diff
public constructor(
     apiUrl: string,
     senderAddress: string,
     signer: SigningCallback | OfflineSigner,
-    enigmautils?: SecretUtils,
-    seed?: Uint8Array,
+    seedOrEnigmaUtils?: Uint8Array | SecretUtils,
     customFees?: Partial<FeeTable>,
     broadcastMode = BroadcastMode.Block,
   ) {
-    super(apiUrl, seed, broadcastMode);
+    if (seedOrEnigmaUtils instanceof Uint8Array) {
+      super(apiUrl, seedOrEnigmaUtils, broadcastMode);
+    } else {
+      super(apiUrl, undefined, broadcastMode);
+    }
+
     this.anyValidAddress = senderAddress;
     this.senderAddress = senderAddress;
     //this.signCallback = signCallback ? signCallback : undefined;
     this.signer = signer;
-    if (enigmautils) {
-      this.restClient.enigmautils = enigmautils;
+    if (seedOrEnigmaUtils && !(seedOrEnigmaUtils instanceof Uint8Array)) {
+      this.restClient.enigmautils = seedOrEnigmaUtils;
     }
```

Thunnini: Looks good to me

assafmo: Thanks, this is already on NPM as `secretjs@0.10.2`

Issue #673, opened 2020-11-20
Handle the new sign doc of the offline signer


Thunnini:
https://github.com/cosmos/cosmjs/blob/v0.23.1/packages/launchpad/src/signingcosmosclient.ts#L93
Currently, the `offlineSigner` interface of `cosmjs` returns the new sign doc that is selected by the user. This new sign doc is neccessary to allow the users to select the gas and gas price that they want. Therefore, the new sign doc should be applied to the result tx because the signature is specific to the new sign doc.

https://github.com/enigmampc/SecretNetwork/blob/dev/v1.0.x/cosmwasm-js/packages/sdk/src/signingcosmwasmclient.ts#L361
However, at this time, secretjs doesn’t handle this and only uses the signature from the offline signer, so the tx signature can’t be verified.

So, I would suggest that the `signAdapter` method returns a new sign doc as well as the signature if the signer is the offline signer. If the signer is a legacy interface, it can return the sign doc as it is. And complete the new tx based on the new sign doc at the final phase.

assafmo:
Fixed by #674

PR #674, opened 2020-11-20
Make the `signAdapter` return `StdTx` with the new `signDoc` applied

Thunnini:
#673 
Currently, the `offlineSigner` interface of `cosmjs` returns the new sign doc that is selected by the user. This new sign doc is necessary to allow the users to select the gas and gas price that they want. Therefore, the new sign doc should be applied to the result tx because the signature is specific to the new sign doc.

The actual `signBytes` that the Keplr or other wallet can have different `fee` or `memo` which have been suggested. So, logic should use the returned `StdSignDoc` as well as a signature to make the transaction.

So, to achieve this, I changed the `signAdapter` return not a signature but `StdTx`. If the signer is a legacy interface, it will use the suggested `SignDoc` as it is.

### DIFF
--- a/cosmwasm-js/packages/sdk/src/signingcosmwasmclient.ts
+++ b/cosmwasm-js/packages/sdk/src/signingcosmwasmclient.ts
@@ -17,6 +17,7 @@ import {
   MsgStoreCode,
   StdFee,
   StdSignature,
+  StdTx,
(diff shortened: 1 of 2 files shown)


assafmo (approved):
Amazing work, man! Thank you. :racing_car:

PR #701, opened 2020-12-17
Update secretcli.md

moonstash:
Is "three" a typo since the docs go on to list four?

If it's a typo then let this update through.

If it's not a typo then what gives?


docs/validators-and-full-nodes/secretcli.md
```diff
@@ -47,7 +47,7 @@
 
 #### Key Types
 
-There are three types of key representations that are used:
+There are four types of key representations that are used:
```
levackt, line 50: There really are 3 types, for background see [cosmos accounts](https://docs.cosmos.network/master/basics/accounts.html) so you can rephrase the section better.

Essentially items 1 and 3 are the same key, it's the address and public key belonging to the same account.

assafmo: Yeah I think there are 3 and someone sent a PR and added another one. We can change that to `four`.

assafmo (approved):

PR #714, opened 2021-01-20
Decrypt output of multiple messages [secretcli]

### LINKED ISSUE #628: Enable decrypting a tx with more than one message using the cli

toml01:
It's not the prettiest thing in the world, but it works.
fixes #628 
 
Example output (**notice how for one of the messages there are errors instead of decrypted text**):
```json
[
  {
    "type": "execute",
    "raw_input": "e9f033de8fa6edaafd1b1c852bfa48172366ca74f9f6d76a0f968f5b115d9728{\"set_deadline\":{\"block\":1}}",
    "input": null,
    "output_data": "",
    "output_data_as_string": "",
    "output_log": [],
    "output_error": {},
    "plaintext_error": "Error: cannot decrypt error: Got an error decrypting the error: siv: authentication failed"
  },
  {
    "type": "execute",
    "raw_input": "cd400fb73f5c99edbc6aab22c2593332b8c9f2ea806bf9b42e3a523f3ad06f62{\"send\":{\"recipient\":\"secret1g6c4vq5me8ymctuvalxsqn38glxrvzadflleck\",\"amount\":\"1231231234564567897897\",\"msg\":\"eyJkZXBvc2l0X3Jld2FyZHMiOnt9fQo=\"}}",
    "input": null,
    "output_data": "",
    "output_data_as_string": "",
    "output_log": [],
    "output_error": {
      "generic_err": {
        "msg": "insufficient funds: balance=18488995000000, required=1231231234564567897897"
      }
    },
    "plaintext_error": ""
  }
]
```

### DIFF
--- a/cosmwasm/contracts/erc20/Cargo.lock
+++ b/cosmwasm/contracts/erc20/Cargo.lock
@@ -330,9 +330,9 @@ checksum = "c36fa947111f5c62a733b652544dd0016a43ce89619538a8ef92724a6f501a20"
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.19"
+version = "1.0.24"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "04f5f085b5d71e2188cb8271e5da0161ad52c3f227a661a3c135fdf28e258b12"
+checksum = "1e0704ee1a7e00d7bb417d0770ea303c1bccbabf0ef1667dae92b5967f5f8a71"
 dependencies = [
  "unicode-xid",
 ]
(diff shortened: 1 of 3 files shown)


assafmo:
@toml01, I don't think you can send multiple messages using the cli, when did we need this?

toml01:
@assafmo You can't, you have to construct it yourself.

This is very useful when you have operations that require multiple transactions.
Specifically for:
* When you want the user to approve only one transaction instead of prompting multiple kleplr windows - for example like in providing liquidity to a new pool
* When you have two transactions that depend on each other, but you don't want one to pass without the other - for example in the original Earn contracts, to top up the rewards you need to 1. send rewards 2. set a new deadline. But if setting the deadline fails, you want to first transaction to fail too.

toml01:
Solved in a nicer way in #1084

PR #717, opened 2021-01-31
added support for debug-print in secret contracts.

reuvenpo:
This PR adds the following:
* When compiled with the `debug-print` feature, the enclave will allow contracts to call the `debug-print` import, which just sends a message to the logger.
* `cosmwasm_std` now exposes a function called `debug_print` and `println!` like macro of the same name that calls it..
* When compiled with the `debug-print` flag, this function will call the `debug_print` export with the contract supplied message. Otherwise, it does nothing.
* The dev image now compiles with this flag.

closes: https://github.com/enigmampc/SecretNetwork/issues/614


assafmo (approved):

levackt:
I'm loving it

Cashmaney (approved):

PR #733, opened 2021-03-18
Update install_cli.md

Steven-Chang:
This link doesn't work anymore

### DIFF
--- a/docs/testnet/install_cli.md
+++ b/docs/testnet/install_cli.md
@@ -38,4 +38,4 @@ chmod +x ./secretcli
 ./secretcli keys add <name>
 ```
 
-See more details on how to use the CLI [here](https://github.com/enigmampc/SecretNetwork/blob/master/docs/secretcli.md)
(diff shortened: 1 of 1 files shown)


reuvenpo:
I think this is a good replacement for the link.  @assafmo is there a better option?
https://github.com/enigmampc/SecretNetwork/blob/master/docs/validators-and-full-nodes/secretcli.md

assafmo:
> I think this is a good replacement for the link. @assafmo is there a better option?
> https://github.com/enigmampc/SecretNetwork/blob/master/docs/validators-and-full-nodes/secretcli.md

Yes. @Steven-Chang can you use this link instead?

Steven-Chang:
@reuvenpo @assafmo - Hey guys, I updated the link as per your suggestions. Let me know if any issues.
Thanks

reuvenpo:
Looks great, thanks @Steven-Chang !

PR #742, opened 2021-04-29
Rumor - light weight node for faster API queries

reuvenpo:
# **WIP**

This work is based on https://github.com/terra-project/mantle-sdk/ and https://github.com/terra-project/mantle/.


assafmo:
@reuvenpo can we close this?

reuvenpo:
Yeah, i don't think this is going to yield any significant progress in the short term. We can revisit it later.

PR #761, opened 2021-06-13
Wasmi performance

reuvenpo:
**This PR adds a cache to instances of `wasmi::Module` instances created in `wasmi-runtime` when executing contracts.**
This single change improved performance by more than 3x.

I additionally optimized hashing the contract code. It would previously hash the entire contract code multiple times in a single interaction, and now happens at most once inside the enclave.

~TODO: figure out the memory overhead of each cached module, and set a limit on the amount of cached modules~ done

UPDATE 2021-08-11: This PR is now superseded by PR #813 and will be closed.


cosmwasm/packages/wasmi-runtime/src/wasm/module_cache.rs
```diff
@@ -0,0 +1,121 @@
...
+    match cache.get(&code_hash).map(create_instance) {
+        Some(Ok(module_ref)) => return Ok(module_ref),
+        None => {} // continue
+
+        // If the stored module failed to process for some reason, remove it.
+        // Shouldn't happen because we already compiled it before.
+        Some(Err(_)) => {
+            cache.remove(&code_hash);
+        }
+    }
```
toml01, line 47: Why do you need this code section here again?

cosmwasm/packages/wasmi-runtime/src/wasm/module_cache.rs
```diff
@@ -0,0 +1,121 @@
...
+use crate::crypto::HASH_SIZE;
+use crate::wasm::types::ContractCode;
+
+use super::gas::{gas_rules, WasmCosts};
+use super::memory::validate_memory;
+use super::runtime::{create_builder, WasmiImportResolver};
+
+lazy_static! {
+    static ref MODULE_CACHE: SgxRwLock<HashMap<[u8; HASH_SIZE], wasmi::Module>> =
+        SgxRwLock::new(HashMap::new());
```
toml01, line 21: What are you using `SgxRwLock` for?

toml01 (approved):

cosmwasm/packages/wasmi-runtime/src/wasm/module_cache.rs
reuvenpo: This code section, as well as the usage or RwLock which you asked about in the other comment, are made with enclave multithreading in mind. So, the RwLock lets you have many concurrent readers, but only one writer. And in this code section, if we were waiting for the write lock, then we check if another thread already wrote to the same location while we were waiting.

check -> lock -> check -> process -> write

cosmwasm/packages/wasmi-runtime/src/wasm/module_cache.rs
reuvenpo: see comment above

reuvenpo:
@toml01 This is still WIP. You approved it but please don't merge this yet 😄

cosmwasm/packages/wasmi-runtime/src/wasm/module_cache.rs
toml01: Yep that's what I thought. Cool :)

toml01:
Merge with failing tests? What kind of savage do you think I am? 😛

reuvenpo:
⚠️ ⚠️ ⚠️ 
The commit above (280036f) fixes an issue i saw during testing, where the enclave's memory was stretched to its limits. It still needs to be tested, to make sure that it doesn't break anything unexpected.

PR #772, opened 2021-07-10
Added a parameter (addedParams) to queryContractSmart

the-dusky:
Updated both cosmwasmclient and restclient, adding addedParams to queryContractSmart allowing flags in query string.


cosmwasm-js/packages/sdk/src/restclient.ts
```diff
@@ -485,13 +485,22 @@ export class RestClient {
...
+
+    /** Loop through addedParams object entries and convert to query string */
+    let paramString = '';
+    if(addedParams) {
+        for (const [key, value] of Object.entries(addedParams)) {
+            paramString = `&${key}=${value}`
+        }
+    }
+
+    const path = `/wasm/contract/${address}/query/${encoded}?encoding=hex${paramString}`;
```
assafmo, line 503: Please use `URLSearchParams` for encoding this.

assafmo (changes requested):
Hi @the-dusky, could you kindly explain why this is need?
Because [this](https://github.com/enigmampc/SecretNetwork/blob/4aacaad3f9659bda2e7a2aa742b9b675ad903e64/x/compute/client/rest/query.go?plain=1#L206-L244) is the server side and I don't see any reference to these query params.

the-dusky:
This is needed to query a contract and provide a "height", which in the CLI would be done using the --height flag. I looked at the code you referenced and admit that I have trouble reading Go. I was however able to test this by editing the code directly in node_modules and it worked, providing the balance for a SNIP-20 at a certain height when connected to an archival node.

One question that I don't know the answer to is: "is 'height' the only param that you can pass anyway?" If so, then the code doesn't need to parse an addedParams object and can instead just take an optional height param. The method I used was a "just in case there are others" kind of thing.

cosmwasm-js/packages/sdk/src/restclient.ts
the-dusky: Made the change

cosmwasm-js/packages/sdk/src/restclient.ts
```diff
@@ -286,7 +286,7 @@ export class RestClient {
   private readonly broadcastMode: BroadcastMode;
   public enigmautils: SecretUtils;
 
-  public codeHashCache: Map<string | number, string>;
+  private codeHashCache: Map<any, string>;
```
assafmo, line 289: Please pull from master.

assafmo:
IDK re `height`, but let's leave it like that. Please just resolve the conflicts with master and then I'll be able to merge this. :blush:

cosmwasm-js/packages/sdk/src/restclient.ts
the-dusky: Sorry about that doing it now

assafmo (approved):

assafmo:
Thanks Sandy! :100:

the-dusky:
Thanks Assaf!

PR #784, opened 2021-07-22
cross compiling works?

Cashmaney:


Makefile
```diff
@@ -116,7 +116,7 @@ build_cli:
 
 xgo_build_secretcli: go.sum
 	@echo "--> WARNING! This builds from origin/$(CURRENT_BRANCH)!"
-	xgo --go latest --targets $(XGO_TARGET) -tags="$(GO_TAGS) secretcli" -ldflags '$(LD_FLAGS)' --branch "$(CURRENT_BRANCH)" github.com/enigmampc/SecretNetwork/cmd/secretcli
+	xgo --image techknowlogick/xgo:go-1.15.x --targets $(XGO_TARGET) -tags="$(GO_TAGS) secretcli" -ldflags '$(LD_FLAGS)' --branch "$(CURRENT_BRANCH)" github.com/enigmampc/SecretNetwork/cmd/secretd
```
assafmo, line 119: I was looking for something like this! :fire:

cmd/secretd/cli_attestation.go
```diff
@@ -0,0 +1,110 @@
...
+	cmd := &cobra.Command{
+		Use:   "init-enclave [output-file]",
+		Short: "Perform remote attestation of the enclave",
+		Long: `Create attestation report, signed by Intel which is used in the registation process of
+the node to the chain. This process, if successful, will output a certificate which is used to authenticate with the 
+blockchain. Writes the certificate in DER format to ~/attestation_cert
+`,
+		Args: cobra.ExactArgs(0),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			println("This is a secretd only function, yo")
```
reuvenpo, line 22: we should probably remove the "yo", yo 😛

go.sum
```diff
@@ -129,6 +129,7 @@ github.com/cosmos/cosmos-sdk v0.40.0-rc1/go.mod h1:4wGruNUDrenXKRl/F7ujW29lTv3C+
 github.com/cosmos/cosmos-sdk v0.40.1 h1:gjrtV3MQj/CMeyXN4+sosHMG6Xwa2uH6HITSjSNL/0E=
 github.com/cosmos/cosmos-sdk v0.40.1/go.mod h1:vlgqdPpUGSxgqSbZea6fjszoLkPKwCuiqSBySLlv4ro=
 github.com/cosmos/cosmos-sdk v0.42.6 h1:ps1QWfvaX6VLNcykA7wzfii/5IwBfYgTIik6NOVDq/c=
+github.com/cosmos/cosmos-sdk v0.42.7 h1:f+ZUjao2y93I37RZ7P2d94JdcEsS7Vq64SBLcNITAVc=
```
reuvenpo, line 132: why are multiple versions of cosmossdk being pulled in?

Makefile
reuvenpo: what does this do, @assafmo

cmd/secretd/cli_attestation.go
assafmo: No, it's obviously necessary

go.sum
assafmo: Go...

Makefile
assafmo: xgo has go 1.13 as its latest version, this^ is the fix

reuvenpo: cool, got it :)

PR #806, opened 2021-08-04
Fixed attestation - syncing nodes now works again

Cashmaney:
Split the dockerfile running tests from the main one to save compilation time and keep images small


x/registration/remote_attestation/remote_attestation.go
```diff
@@ -135,6 +131,10 @@ func verifyCert(payload []byte) ([]byte, error) {
 
 	opts := x509.VerifyOptions{
 		Roots: roots,
+		// note: there's no way to not validate the time, and we don't want to write this code
+		// ourselves. We also can't just ignore the error message, since that means that the rest of
+		// the validation didn't happen (time is validated early on)
+		CurrentTime: time.Date(2023, 11, 04, 00, 00, 00, 00, time.UTC),
```
assafmo, line 137: What does this mean?

Cashmaney: Remember the 90 day certificate thing from a while back? It's the fix for that

Cashmaney: I merged a few older PRs here too, for good measure

PR #807, opened 2021-08-05
Migrate secretjs to Stargate + miscellaneous CLI fixes 

Cashmaney:


assafmo:
- migrate secretjs to Stargate (still untested)
- fix default gas for queries
- fix --from on "compute execute"
- make query errors more verbose
- fix default app.toml config, also on our docker images
- fix decryption of execute output data

PR #816, opened 2021-08-18
update secret-contract-optimizer rust version to 1.54

darwinzer0:
Updates the secret-contract-optimizer to use Rust v 1.54. 

The newer version of Rust prevents `const generics are unstable` error when compiling some third-party libraries. For more on the change see: https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html#const-generics-mvp


reuvenpo:
Thanks for the PR Darwin!
I'll merge this now and we'll build and upload a new version to dockerhub a little later 🙂

PR #821, opened 2021-08-25
Update setup-sgx node doc with newest Intel SGX drivers

dylanschultzie:
Updates the setup process with the newest Intel SGX drivers, unless the node is on ubuntu 16.04, in which case Intel is no longer supporting them. 

Tested the install on Ubuntu versions 16/18/20.

Nodefather (I'm not sure of their non-discord handle) said they had to do extra command to get Ubuntu 20 working, but I was unable to replicate. The commands are as follows:

```
sudo apt update
sudo apt install build-essential
sudo apt install sgx-aesm-service libsgx-uae-service libsgx-launch libsgx-urts
```

Signed-off-by: Dylan Schultz <dylanschultz311@gmail.com>

### DIFF
--- a/docs/validators-and-full-nodes/setup-sgx.md
+++ b/docs/validators-and-full-nodes/setup-sgx.md
@@ -13,7 +13,7 @@ If you're running a local machine and not a cloud-based VM -
 
 ### Install SGX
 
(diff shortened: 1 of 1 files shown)


assafmo (approved):
Thanks @dylanschultzie! :100:

PR #832, opened 2021-09-27
Remove Validator docs

dylanschultzie:
This is paired with: https://github.com/SecretFoundation/docs/pull/13

Basically, there have been recurring question around `build.scrt.network` and `docs.scrt.network` as they share similar content. I'm pruning it down so the content is unique across them.


assafmo:
Hi @dylanschultzie, we're going to discuss this internally before we'll merge this.

dylanschultzie:
> Hi @dylanschultzie, we're going to discuss this internally before we'll merge this.

For sure, I figured as much. At the very least it'd be nice to take down the Validator section of build.scrt.network so that information wasn't duplicated. 

As a quick example of the confusion, the `secret agent` program asks what the slashing rate is for validators. build.scrt.network says 1%, but docs.scrt.network says 0.01% (which is the correct number).

assafmo:
Thanks @dylanschultzie, we'll start writing docs in the https://github.com/SecretFoundation/docs. I also think we can kill https://build.scrt.network and add a build section in https://docs.scrt.network. No reason to keep them separate.

Please give us (@assafmo, @reuvenpo, @toml01 & @Cashmaney) write access to https://github.com/SecretFoundation/docs cc @moonstash.

PR #850, opened 2021-11-18
WIP: Put enclave access behind a reentrant mutex

reuvenpo:
This PR will fix the issue of queries failing with `SGX_ERROR_OUT_OF_TCS` when another query is already running at the same time.
Now, queries will wait for up to 2 seconds to enter the enclave before failing with an error message that says `"Too many queries, please try again"`.

### DIFF
--- a/cosmwasm/packages/sgx-vm/src/attestation.rs
+++ b/cosmwasm/packages/sgx-vm/src/attestation.rs
@@ -127,7 +127,7 @@ pub extern "C" fn ocall_get_update_info(
 }
 
 pub fn create_attestation_report_u(spid: &[u8], api_key: &[u8]) -> SgxResult<()> {
-    let enclave = get_enclave()?;
+    let enclave = get_enclave()?.ok_or(sgx_status_t::SGX_ERROR_OUT_OF_TCS)?;
 
     let eid = enclave.geteid();
     let mut retval = sgx_status_t::SGX_SUCCESS;
@@ -156,7 +156,7 @@ pub fn create_attestation_report_u(spid: &[u8], api_key: &[u8]) -> SgxResult<()>
 pub fn untrusted_get_encrypted_seed(
     cert: &[u8],
 ) -> SgxResult<Result<[u8; ENCRYPTED_SEED_SIZE], NodeAuthResult>> {
-    let enclave = get_enclave()?;
+    let enclave = get_enclave()?.ok_or(sgx_status_t::SGX_ERROR_OUT_OF_TCS)?;
     let eid = enclave.geteid();
     let mut retval = NodeAuthResult::Success;
     let mut seed = [0u8; ENCRYPTED_SEED_SIZE];
--- a/cosmwasm/packages/sgx-vm/src/enclave.rs
+++ b/cosmwasm/packages/sgx-vm/src/enclave.rs
@@ -1,6 +1,9 @@
-use std::{env, path::Path, sync::Mutex};
+use enclave_ffi_types::{EnclaveBuffer, RuntimeConfiguration};
+use std::env;
+use std::ops::Deref;
+use std::path::Path;
+use std::time::Duration;
 
-use enclave_ffi_types::RuntimeConfiguration;
 use sgx_types::{
     sgx_attributes_t, sgx_enclave_id_t, sgx_launch_token_t, sgx_misc_attribute_t, sgx_status_t,
     SgxResult,
@@ -9,6 +12,9 @@ use sgx_urts::SgxEnclave;
 
 use lazy_static::lazy_static;
 use log::*;
+use parking_lot::{Mutex, ReentrantMutex, ReentrantMutexGuard};
+
+use crate::wasmi::imports;
 
 static ENCLAVE_FILE: &str = "librust_cosmwasm_enclave.signed.so";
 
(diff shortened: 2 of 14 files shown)


cosmwasm/packages/sgx-vm/src/attestation.rs
```diff
@@ -127,7 +127,7 @@ pub extern "C" fn ocall_get_update_info(
 }
 
 pub fn create_attestation_report_u(spid: &[u8], api_key: &[u8]) -> SgxResult<()> {
-    let enclave = get_enclave()?;
+    let enclave = get_enclave()?.ok_or(sgx_status_t::SGX_ERROR_OUT_OF_TCS)?;
```
reuvenpo, line 131: when `get_enclave` returns `Ok(None)` that means that the enclave is busy. In most code contexts, such as this one, it's in calls to the enclave while the node is still being configured. We don't expect this to really happen but because i preferred to not panic, i just convert it to the original error.

cosmwasm/packages/sgx-vm/src/enclave.rs
```diff
@@ -125,3 +163,30 @@ pub fn configure_enclave(config: EnclaveRuntimeConfig) -> SgxResult<()> {
 
     Ok(())
 }
+
+/// This is a safe wrapper for allocating buffers inside the enclave.
+///
+/// It must be called after the enclave has been initialized, and can not be called
+/// while another thread is using the enclave, or it will panic.
+pub(super) fn allocate_enclave_buffer(buffer: &[u8]) -> SgxResult<EnclaveBuffer> {
```
reuvenpo, line 171: I moved this function here just to clean up the code a little. It was in a really inappropriate location before

cosmwasm/packages/sgx-vm/src/errors/enclave.rs
```diff
@@ -15,6 +15,8 @@ pub enum EnclaveError {
         status: sgx_types::sgx_status_t,
         backtrace: Backtrace,
     },
+    #[snafu(display("Too many queries, please try again"))]
+    EnclaveBusy {},
 }
```
reuvenpo, line 20: Users will see this string if a query times out on enclave access. Is this ok?

cosmwasm/packages/sgx-vm/src/instance.rs
```diff
@@ -87,7 +87,9 @@ where
         let module = compile(code)?;
         Instance::from_module(&module, deps, gas_limit)
         */
-        let enclave = get_enclave().map_err(EnclaveError::sdk_err)?;
+        let enclave = get_enclave()
+            .map_err(EnclaveError::sdk_err)?
+            .ok_or(EnclaveError::EnclaveBusy {})?;
         let module = Module::<S, Q>::new(
```
reuvenpo, line 93: This is the context where queries try to access the enclave. If it's busy (`Ok(None)` was returned) then we tell the users about it.

x/compute/client/utils/utils.go
```diff
@@ -252,7 +252,7 @@ var re = regexp.MustCompile("encrypted: (.+?):")
 func (ctx WASMContext) DecryptError(errString string, msgType string, nonce []byte) (json.RawMessage, error) {
 	regexMatch := re.FindStringSubmatch(errString)
 	if len(regexMatch) != 2 {
-		return nil, fmt.Errorf("Got an error finding base64 of the error: regexMatch '%v' should have a length of 2", regexMatch)
+		return nil, fmt.Errorf("Got an error finding base64 of the error: regexMatch '%v' should have a length of 2. error: %v", regexMatch, errString)
 	}
```
reuvenpo, line 256: This is the code that prints an error when the query fails. I changed this in order to be able to see the TCS message in the CLI in my testing.

reuvenpo (commented):
Notes to reviewer:

cosmwasm/packages/sgx-vm/src/errors/enclave.rs
Cashmaney: Better to do "Too many queries, node is processing block" - more informative for node runners

cosmwasm/packages/sgx-vm/src/enclave.rs
```diff
@@ -68,16 +102,20 @@ fn init_enclave() -> SgxResult<SgxEnclave> {
...
 
 /// Use this method when trying to get access to the enclave.
 /// You can unwrap the result when you are certain that the enclave
 /// must have been initialized if you even reached that point in the code.
-pub fn get_enclave() -> SgxResult<&'static SgxEnclave> {
-    SGX_ENCLAVE.as_ref().map_err(|status| *status)
+/// If `Ok(None)` is returned, that means that the enclave is currently busy.
+pub fn get_enclave() -> SgxResult<Option<EnclaveGuard>> {
+    let mutex = SGX_ENCLAVE_MUTEX.as_ref().map_err(|status| *status)?;
+    let maybe_guard = mutex.get_enclave(Duration::from_secs(6));
```
Cashmaney, line 117: Make this a const instead of a magic number, even if it isn't fully customizable

cosmwasm/packages/sgx-vm/src/enclave.rs
```diff
@@ -101,14 +139,15 @@ impl EnclaveRuntimeConfig {
 }
 
 pub fn configure_enclave(config: EnclaveRuntimeConfig) -> SgxResult<()> {
-    let mut configured = SGX_ENCLAVE_CONFIGURED.lock().unwrap();
+    let mut configured = SGX_ENCLAVE_CONFIGURED.lock();
```
Cashmaney, line 142: Seems kind of strange to have this sort of pattern around a config, but I'm assuming this isn't part of the PR so w/e

cosmwasm/packages/sgx-vm/src/errors/enclave.rs
reuvenpo: it's not really related to blocks. A query won't be waiting there if a block is running

cosmwasm/packages/sgx-vm/src/enclave.rs
reuvenpo: I did this in case this was somehow called from multiple threads. it should only be called once in the node init though so I'm not sure why i thought that was necessary.

cosmwasm/packages/sgx-vm/src/enclave.rs
reuvenpo: done 👍🏻

PR #857, opened 2021-11-29
Speed up queries

reuvenpo:
This PR will implement various techniques to speed up queries for API nodes.

implemented so far:
* Skip floating point check in contract queries


assafmo:
Queries can still happen inside of transactions, so I think the better solution is to do floating point checks in store or init and skip it in handle and query.

faddat:
use rocksdb #868 -- it helps, and I'd love your feedback

assafmo:
Hi @faddat, Thanks for your feedback, we'll relay it to our node runners 😇

PR #859, opened 2021-12-05
Tendermint rollback command (rolls back state by 1 block)

Cashmaney:
Still need to test it properly


Cashmaney:
Doesn't quite work yet - need to plug in the sdk for this to work properly

PR #861, opened 2021-12-12
Enclave multithreading

reuvenpo:
This PR is based on and supersedes #531.
The purpose of this branch is to experiment with creating a query-optimized enclave that can handle multiple queries running in 16 parallel threads.


cosmwasm/packages/sgx-vm/src/enclave.rs
```diff
@@ -169,29 +128,82 @@ pub fn configure_enclave(config: EnclaveRuntimeConfig) -> SgxResult<()> {
...
+    fn drop(&mut self) {
+        // eprintln!("Query Token destruction. recursive: {}", self.recursive);
+        if !self.recursive {
+            let mut count = self.doorbell.count.lock();
+            // eprintln!(
+            //     "The current count of tasks is {}/{}, attempting to decrease.",
+            //     TCS_NUM - *count,
+            //     TCS_NUM
+            // );
+            *count += 1;
```
toml01, line 200: Shouldn't it be `-= 1` ?

cosmwasm/packages/sgx-vm/src/wasmi/wrapper.rs
```diff
@@ -4,22 +4,47 @@ use std::ffi::c_void;
...
-use crate::enclave::EnclaveGuard;
+
+/// This is a safe wrapper for allocating buffers inside the enclave.
+pub(super) fn allocate_enclave_buffer(buffer: &[u8]) -> SgxResult<EnclaveBuffer> {
+    let ptr = buffer.as_ptr();
+    let len = buffer.len();
+    let mut enclave_buffer = EnclaveBuffer::default();
+
+    let enclave_id = crate::enclave::get_enclave()
+        .expect("If we got here, surely the enclave has been loaded")
```
toml01, line 33: I think this message should be changed, because now the enclave in not Mutex protected?

toml01 (commented):
On a general note - is there something that prevents multiple handles to happen concurrently? i.e. if I'm not mistaken, we want to allow only queries to use multiple threads, while allowing only one concurrent init/handle

cosmwasm/packages/sgx-vm/src/enclave.rs
```diff
@@ -125,3 +127,69 @@ pub fn configure_enclave(config: EnclaveRuntimeConfig) -> SgxResult<()> {
 
     Ok(())
 }
+
+/// This const determines how many seconds we wait when trying to get access to the enclave
+/// before giving up.
+const ENCLAVE_LOCK_TIMEOUT: u64 = 6;
+const TCS_NUM: u8 = 16;
```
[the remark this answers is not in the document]
reuvenpo: Maybe... It would require XML parsing, but I'll check

cosmwasm/packages/sgx-vm/src/enclave.rs
```diff
@@ -130,14 +130,15 @@ pub fn configure_enclave(config: EnclaveRuntimeConfig) -> SgxResult<()> {
 
 /// This const determines how many seconds we wait when trying to get access to the enclave
 /// before giving up.
-const ENCLAVE_LOCK_TIMEOUT: u64 = 6;
+const ENCLAVE_LOCK_TIMEOUT: u64 = 6 * 5;
```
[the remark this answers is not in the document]
reuvenpo: Yeah, I did this just for testing purposes, but thanks for the reminder!

cosmwasm/packages/sgx-vm/src/enclave.rs
reuvenpo: No, the count represents the amount of "free slots" that are available in the enclave runtime. So when the query token is destroyed, we want to release the slot, by saying there's now one more available.

reuvenpo:
> is there something that prevents multiple handles to happen concurrently
Yeah, there's a RWLock in the cosmos SDK, so it won't send more than one handle down here, which is why i don't check for that.

cosmwasm/packages/sgx-vm/src/enclave.rs
toml01: Ahh right right, my bad

PR #871, opened 2022-01-09
Rocksdb test

Cashmaney:
Leaving this here for now - adding a temp tag to compile automatically


faddat:
FYI, personally I don't recommend the .deb distribution, instead distribute easy directions on building from source to validators.  I'll keep you informed.

assafmo:
@faddat to be able to build for production, one must register for Intel API keys. We want to spare this from validators, but they can choose to do so. We don't publish our (SCRT Labs') keys to prevent a black hat from misusing them and making Intel to blacklist the keys, which will cause all nodes registrations with our build to fail.

To build for development purposes you can use empty keys.

assafmo:
Seems like that change from cosmos-sdk v0.44.3 to v0.44.5 broke it, as we made a few changes in our fork and now the `replace` at the bottom of `go.mod` doesn't work.

faddat:
https://github.com/scrtlabs/SecretHoldEm

really justifies a lot.

Cashmaney:
@faddat The process you describe is part of the attestation process that happens with Intel and that SGX trust is based on. SGX encalves are signed with keys whitelisted by Intel. Only enclaves signed by those keys can run in production. This happens during compilation.

At runtime each node/enclave goes through the attestation process with Intel, that verifies the patch level and platform the enclave is running on. To go through this attestation process a different set of keys (api key and SPID key) are required to communicate with Intel's attestation service - these are not the keys used to sign the enclave, and are bundled together with the runtime for simplicity.

Finally, the node performs an on-chain transaction broadcasting their attestation results to the network. The network then checks that the registering node matches the exact code and signature that is currently being ran by all the other nodes.

As you can see, this process is slightly more complex than a standard chain so hopefully this made a bit of sense:)

Cashmaney:
Closing this in favor of #884

PR #873, opened 2022-01-13
Crates io publishing

reuvenpo:
I published the latest version of our cosmasm-std to crates.io here:

https://crates.io/crates/secret-cosmwasm-std

To use it, you can import it like so:
```toml
cosmwasm-std = { package = "secret-cosmwasm-std", version = "0.10.0" }
```

The latest commit here is also tagged as `secret-cosmwasm-v0.10.0`

Coming to `secret-toolkit` soon!


reuvenpo:
These changes have been used successfully by users for about a week now so I merged it to master

PR #903, opened 2022-03-13
issues 874, 875, 876

kent-3:
Modified the `genesis.json` file to replace the 172800s (2 days) voting period with 90s.
Added mnemonic phrases to be used for the genesis accounts, so they are always the same.
Initialized accounts c & d with funds.

I don't know how to actually build the docker image to test it fully, but these changes _should_ address the three issues mentioned in the title. The `echo $a_mnemonic | secretd keys add a --recover` method works in the command line.


assafmo:
Thank you @kent-3!

Issue #918, opened 2022-04-12
Add ARM targets to the secretcli release CI 


Cashmaney:
Right now the CI automatically compiles secretcli for amd64 windows/mac/linux. We should add M1 Mac support as there are more and more apple fanboys out there 😛

assafmo:
This might be a good place to start https://github.com/scrtlabs/SecretNetwork/blob/3a2e013f57b7bb51b7e82a62ea9ce11544dfd1a7/Makefile#L160-L162

assafmo:
Cc @faddat

PR #922, opened 2022-04-26
arm64 build targets for released cli

faddat:
* add linux-arm64
* add darwin-arm64

closes: #918

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -135,6 +135,7 @@ jobs:
         run: |
           make build_cli
           cp "secretcli" "secretcli-$RUNNER_OS"
+          make build_linux_arm64_cli
       - uses: actions/upload-artifact@v2
         with:
           name: secretcli-${{runner.os}}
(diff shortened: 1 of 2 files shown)


faddat:
Hi, I think you may need to enable CI for this.

faddat:
https://dev.azure.com/enigmampc/Secret%20Network/_build/results?buildId=2566&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=5caf77c8-9b10-50ef-b5c7-ca89c63e1c86&l=83

assafmo:
Hi @faddat, thanks for this!
I think that the CI failing is not related to this PR.

Cashmaney:
Yeah, the CI will not run properly on targets coming from forks, but it should be all good. We can merge this and add the target to the runner matrix

.github/workflows/release.yaml
```diff
@@ -135,6 +135,7 @@ jobs:
         run: |
           make build_cli
           cp "secretcli" "secretcli-$RUNNER_OS"
+          make build_linux_arm64_cli
```
Cashmaney, line 138: Copying from the other comment:

The arm64 build was added to a line that runs on each build target in the matrix. A better way is to add the target to the matrix instead, that way it gets built and uploaded/released automatically. IMO linux-arm64 is supported, and there should be a mac m1 (darwin-arm64) target supported on GH actions, though I'm not 100% about that

Cashmaney:
Closing this in favor of #925

PR #926, opened 2022-05-03
New registration service

Cashmaney:
Moved registration service to target new endpoints, and refactored handling.

End result is that the user now only specifies the registration service and does not directly interact with a node during init. In addition, there are defaults for testnet and mainnet with an option to specify a custom service for future testnets/upgrades


assafmo (approved):
LGTM LFG 🚀

PR #932, opened 2022-05-22
Update math.rs

kent-3:
Added a subtraction method for the Decimal type. I realize the math module will be getting overhauled with CosmWasm v1, but this could be useful to someone in the meantime.


Cashmaney:
LGTM, thanks for the contribution

assafmo (commented):
Thanks @kent-3! :chart_with_upwards_trend:

PR #941, opened 2022-06-12
LocalSecret: Cross compile to ARM64 (Mac M1)

assafmo:


assafmo:
https://github.com/apache/incubator-teaclave-sgx-sdk/issues/318#issuecomment-998521315

PR #943, opened 2022-06-23
Query without gas metering

assafmo:
This PR increases query performance by about 25% (1.25x).

| A large query | Time (s) |
|--|--|
| Without gas metering |  41.273849005 |
| With gas metering |     52.79072004 |

Notes:
- This cannot be released as a non-breaking v1.3.x version. A contract can use `query_chain()` to force different gas behavior from current v1.3.x versions and thus break consensus.
- TODO: Test like hell that this cannot happen inside a tx.
- TODO: On CosmWasm v1 make sure that it's not gameable using `StaregateQuery`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,7 +4,7 @@ on: [push]
 
 jobs:
   Enclave-Unit-Tests:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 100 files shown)


assafmo:
Closing. Interesting results but this version of the code won't see the light of day (wasm3 is coming then wasmer).

PR #946, opened 2022-07-03
gofumpt -w -l .

faddat:
This PR changes nothing except for formatting. 

It'll make the code more attractive-- it's designed to be merged before the upcoming linter PR, which will start with this PR as its base because working with fumpted code is easier.


assafmo (approved):
Gucci

PR #947, opened 2022-07-03
go 1.18

faddat:
The goal of this PR is to bring us closer to the package selection that will be used on developer laptops.


faddat:
thanks so much for the ci fixes on this and #946

assafmo (approved):
Slap

PR #956, opened 2022-07-10
wasm3

reuvenpo:


Cashmaney:
closing this - will reopen with the wasm3-cw1 branch so we don't maintain multiple versions

PR #962, opened 2022-07-13
Test v1 BankMsg::Burn & v0.10+v1 CosmosMsg::Custom

assafmo:


assafmo:
Essentially inside the enclave CustomMsg was defined like [this](https://github.com/scrtlabs/SecretNetwork/blob/6ce26f98/cosmwasm/enclaves/shared/cosmwasm-types/src/types.rs#L177-L190 ) to make [these tests](https://github.com/scrtlabs/SecretNetwork/blob/a6071fa3/x/compute/internal/keeper/reflect_test.go) pass, but in practice CustomMsg should always fail [hare](https://github.com/scrtlabs/SecretNetwork/blob/a6071fa30db4e9989ac0fa0dbf8f0bda34f0e956/x/compute/internal/keeper/handler_plugin.go#L149-L151) (v0.10) or [here](https://github.com/scrtlabs/SecretNetwork/blob/a6071fa3/x/compute/internal/keeper/handler_plugin.go#L149-L151) (v1).

PR #964, opened 2022-07-14
Create codeql.yml

faddat:
Hey I think that this may need to run on a GR or we need to give it special permission to run on github actions but it's really really great.

It detects things like non-determinism errors.


assafmo:
Is this static analysis magic?

faddat:
it is exactly the static analysis magic

Cashmaney:
Very cool. Interested to see whether it will work out of the box, or if we'll need to tinker with it

faddat:
I find that setting the tag secretcli can help, locally.  But also seems my local dev env "ain't quite right".  

When I set the build tag, it works better but it doesn't seem to work perfectly.

Cashmaney:
> I find that setting the tag secretcli can help, locally. But also seems my local dev env "ain't quite right".
> 
> When I set the build tag, it works better but (eg: gopls) doesn't seem to work perfectly.

The secretcli tag just compiles without the SGX enclave part, so it's not really a complete build. 

Also, are we missing anything else for this action to run? I was expecting the codeql workflow to be available in this branch already 🤔

assafmo:
#1020

PR #966, opened 2022-07-14
WIP: CosmWasm v1 IBC Receive Hooks

assafmo:


x/compute/internal/keeper/keeper.go
```diff
@@ -1088,25 +1066,27 @@ func (k Keeper) reply(ctx sdk.Context, contractAddress sdk.AccAddress, reply v1w
 	consumeGas(ctx, gasUsed)
 
 	if execErr != nil {
-		return nil, sdkerrors.Wrap(types.ErrReplyFailed, execErr.Error())
```
liorbond, line 1091: @assafmo why?

PR #968, opened 2022-07-15
Goimports v1

faddat:
Adds a goimports to formatting.  This changed some .pb.go files but I feel like allowing goimports to do that is easier than excluding them: everything can be run with defaults.


assafmo:
Isn't gofumt makes goimports obsolete?

faddat:
correct.

assafmo:
So do we still need this PR?

PR #969, opened 2022-07-15
Gitignore

faddat:
* adds /.vscode to the list of ignored files & paths
* improves default vscode settings

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -20,7 +20,6 @@
   "go.useLanguageServer": true,
   "go.lintTool": "golangci-lint",
   "go.lintOnSave": "workspace",
-  "go.buildTags": "secretcli",
   "gopls": {
     "formatting.gofumpt": true
   },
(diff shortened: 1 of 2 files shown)


assafmo:
How are these two changes working together?

faddat:
Ah, they work fine together.

Here's how it works:

.gitignore tells git not to update those files

if we want to change the default dev env settings, then we 

git add .vscode/settings.json --force
git commit -m "update vscode settings"

.vscode/settings.json
```diff
@@ -18,6 +18,9 @@
     "RUST_BACKTRACE": "1"
   },
   "go.useLanguageServer": true,
+  "go.lintTool": "golangci-lint",
+  "go.lintOnSave": "workspace",
+  "go.buildTags": "secretcli",
```
assafmo, line 23: So this actually messes up debug mode, since it tells it to not connect to the rust code. Can you remove this line?

assafmo:
> git add .vscode/settings.json --force

Didn't know about this feature. Cool.

PR #970, opened 2022-07-19
Custom chain ID from env

levackt:
CHAINID via env var doesn't work, it's always secretdev-1

### DIFF
--- a/deployment/docker/bootstrap/bootstrap_init.sh
+++ b/deployment/docker/bootstrap/bootstrap_init.sh
@@ -6,11 +6,7 @@ if [ ! -e "$file" ]; then
   rm -rf ~/.secretd/*
   rm -rf /opt/secret/.sgx_secrets/*
 
-  if [ -z "${CHAINID}" ]; then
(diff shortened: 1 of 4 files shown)


levackt:
The problem is the logic is inverted, so if the shorthand in the PR isn't preferred, we can fix it like so;
```sh

if [[ -z "${CHAINID}" ]]; then
    chain_id="secretdev-1"
else
    chain_id=$CHAINID
fi
```

levackt:
Also present [here](https://github.com/scrtlabs/SecretNetwork/blob/88ccc92433ef3f5149226705e4a26a3c2ebc2279/deployment/docker/bootstrap/bootstrap_init.sh#L9), but probably only used by CI with defaults

assafmo:
Cool, didn't know of this syntax.

levackt:
Sure, let me check

levackt:
Tests look better, squashed the commits and undrafting, should build

Cashmaney:
LGTM! Thanks @levackt, merging this now

PR #1020, opened 2022-07-20
Create codeql-analysis.yml

Cashmaney:


faddat:
This is a really good one imo. 

Makes the scary things stay away.

Issue #1024, opened 2022-07-23
Upgrade rust?


faddat:
Ok, so first rust for me. 

I have a build of all of the go components for v1 working, but I cheated. 

Some thing that has always concerned me about CW is that we use old versions of rust but for the wrong reasons (gas pricing consistency eg: don't do that, it would then go too fast) if I understand correctly, there's a pretty significant reimplementation of CW living in this repository.

And we build it with a 2020 vintage compiler, because of SGX.  I'm wondering if we can get a fresher compiler going-- maybe 2022 era? 

If it is just a matter of a applying an SGX patch, I'd love to give it a go, but I kinda figure that there are big performance gains to be had by modernizing the rust build toolchain in scrt. 

**update**

I installed the nightly rust toolchain, and I removed the rust-toolchain file in the repo root.

Stuff started working like magic.

Cashmaney:
It's actually a bit more complicated than that. Because a lot of the code depends on [teaclave-sgx-sdk](https://github.com/apache/incubator-teaclave-sgx-sdk) we have to use a very specific version of Rust for most of the code. The annoying part is that this applies not only for the sgx-sdk itself, but inside the enclave a lot of the packages are forks that were specifically modified to work with a specific sgx-sdk version. 

I'd like nothing more than to use a version of Rust that is newer, but to do it right now would probably require forking and updating like 20/30 different repos. That's why we're looking at either using [version 2.0 of teaclave-sgx-sdk](https://github.com/apache/incubator-teaclave-sgx-sdk/tree/v2.0.0-preview) or porting our code over to Gramine - both of which will be far, far easier to maintain

PR #1026, opened 2022-07-24
Faddat/tiny changes slowly

faddat:


.cargo.toml
```diff
@@ -0,0 +1,9 @@
+# This is appended to the build system's .cargo/cargo.toml file
+
+[source."https://github.com/apache/teaclave-sgx-sdk.git"]
+git = "https://github.com/apache/teaclave-sgx-sdk.git"
+rev = "a37ffb9449ba6d5b6e4a9d586bbab864ae732269"
+replace-with = "vendored-sources"
+
+[source.vendored-sources]
+directory = "third_party/vendor"
```
reuvenpo, line 9: Can you please explain what this does? 

We don't want this sort of thing to be applied to all the rust modules in this repo, only the enclave modules.

cosmwasm/enclaves/rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2020-10-25
+nightly
```
reuvenpo, line 1: We usually want to pin the enclave's compiler version, because:
- Teaclave only explicitly support specific nightly versions.
- We want to make sure our build are relatively reproducible, and we all see the same compiler bugs if they pop up.
- This also makes sure we all work with the same language features.

rust-toolchain
```diff
@@ -1 +1 @@
-1.54
\ No newline at end of file
+nightly
```
reuvenpo: This should be a stable version. This file applies the version to go-cosmwasm (the rust part), sgx-vm, and the example contracts, all of which do not require nightly.

reuvenpo (changes requested):
Thanks for the PR, but please see my comments below on toolchain versioning. I don't think we should be changing it as you have. As far as we understand it, this is the correct approach:

To update the nightly compiler used for the enclaves, we'll also need to update the teaclave version we're using as well as a bunch of other custom dependency overrides in the enclaves that are provided by the teaclave project. Some dependencies are our own custom forks which will need to be handled separately.

Maybe the top-level rust-toolchain file can be updated to 1.62.1.

CC: #1024

EDIT: Itzik comment [here](https://github.com/scrtlabs/SecretNetwork/issues/1024#issuecomment-1193270748) explains an even better approach

cosmwasm/enclaves/rust-toolchain
faddat: aha!

this is key info

faddat:
its mainly me playing with rust

PR #1064, opened 2022-08-08
Faddat/go 19

faddat:
Go 1.19 contains some cryptographic tweaks and it seems to be a good idea to be early adopters.

https://tip.golang.org/doc/go1.19

Changes:

* golangci-lint with reasonable exclusions to avoid rewriting scrt
* use io and os instead of io/ioutil so that we don't trigger go 1.19's anger
* basic linting changes
* fix GetTXConfig
* Now we can incrementally adopt tighter linting


faddat:
I think that the remaining error **might** be a build system error. 

Not certain.

faddat:
It isn't the build system.  I used the same temp directory for all the things.

faddat:
good to go :)

assafmo (approved):

assafmo:
Thank you! This looks good, however to prevent future headache when merging cosmwasm-v1 into master, I'm closing this PR and I've cherry-picked the commits (a9a2e83, 5a8676f, c157695 & c94929e) into cosmwasm-v1.

PR #1066, opened 2022-08-09
stop using ioutil

faddat:
Ioutil has been deprecated for quite some time, and we should not use it.


assafmo (approved):
Thank you, @faddat!

PR #1072, opened 2022-08-15
Removed all teaclave usages. Currently only check-enclave flow functions

eladr7:
Notice that the repo has no query/execute functionalities.
These should be implemented on the gramine side.


Cashmaney:
From a quick overview - 
1. Can you restore third_party/proto? It is used in multiple places
2. Can you add the gramine server to this repo? It will be easier to track if it's all in one place

PR #1080, opened 2022-08-21
Add all of the missing RPC apis

liorbond:


proto/secret/compute/v1beta1/query.proto
```diff
@@ -34,10 +34,24 @@ service Query {
...
+    // Query contract hash by address
+    rpc ContractHash (QueryContractHashRequest) returns (QueryContractHashResponse) {
+        option (google.api.http).get = "/compute/v1beta1/hash/address/{address}";
+    }
     // Query contract hash by code id
     rpc ContractHashByID (QueryContractHashByIDRequest) returns (QueryContractHashResponse) {
         option (google.api.http).get = "/compute/v1beta1/hash/id/{code_id}";
     }
+    // Query contract key by address
+    rpc ContractKey(QueryContractKeyRequest) returns (QueryContractKeyResponse) {
```
assafmo, line 46: Irrelevant here

proto/secret/compute/v1beta1/query.proto
```diff
@@ -34,10 +34,24 @@ service Query {
...
+        option (google.api.http).get = "/compute/v1beta1/key/{address}";
+    }
+    // Query contract address by label
+    rpc AddressByLabel(QueryContractAddressByLabelRequest) returns (QueryContractAddressByLabelResponse) {
```
assafmo, line 50: Should also add the other way around

proto/secret/compute/v1beta1/query.proto
liorbond: Why? it was implemented in the querier but was missed here, If it's irrelevant ill remove it from the querier as well.

proto/secret/compute/v1beta1/query.proto
liorbond: Working on it in another PR, this PR focus on those APIs that were implemented in the queried but weren't exported by the pb

Cashmaney (approved):

proto/secret/compute/v1beta1/query.proto
assafmo: Because it's used only inside the enclave, there's no reason to use it outside

assafmo:
Replaced by https://github.com/scrtlabs/SecretNetwork/pull/1130

PR #1081, opened 2022-08-21
Add missing RPC function and make proto file readable

liorbond:


Cashmaney (changes requested):
When using addresses as "github.com/cosmos/cosmos-sdk/types.AccAddress" they are encoded in url unsafe base64 and cannot be used safely in the query - some will fail, and the others will be uncoded in a weird way. Use "string" instead to get the bech32 encoded address that we're looking for

Cashmaney (approved):

assafmo:
Replaced by https://github.com/scrtlabs/SecretNetwork/pull/1130

PR #1084, opened 2022-08-22
Decrypt multiple messages in a single TX

### LINKED ISSUE #628: Enable decrypting a tx with more than one message using the cli

toml01:
This might break some tests. Should not merge unless all tests pass.

Fixes #628


toml01:
Replaced with https://github.com/scrtlabs/SecretNetwork/pull/1085

PR #1085, opened 2022-08-22
Decrypt multiple messages in a single TX

### LINKED ISSUE #628: Enable decrypting a tx with more than one message using the cli

toml01:
This might break some tests. Should not merge unless all tests pass.

Fixes #628


Cashmaney (dismissed):

.gitignore
```diff
@@ -25,6 +25,7 @@ SHA256SUMS*
 __debug_bin
 id_tx_io.json
 .idea
+.vscode
```
liorbond, line 28: why?
we are changing this to make vscode developers life better :P

toml01: hmm because why would we commit stuff like `dryrun.log` and `configurationCache.log`?
if we want we can upload `launch.json` specifically

liorbond: Maybe we need to ignore .vscode/*.log

assafmo: Then gitignore `dryrun.log` * `configurationCache.log`

assafmo: We've added debug configs, linting and formatting for go, and rust-analyzer config for the enclave so now it's 1000x better than CLion.

liorbond (approved):

.gitignore
toml01: also lol:
https://github.com/scrtlabs/SecretNetwork/pull/1085/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947L22

liorbond (approved):

toml01:
#1135

PR #1086, opened 2022-08-23
Limit length of error messages

Cashmaney:


assafmo (approved):
🔥

toml01:
Might wanna merge into `master` since the v1 branch doesn't pass tests currently

PR #1087, opened 2022-08-23
Rust logs `stdout` -> `stderr`

### LINKED ISSUE #942: 'secretd export' prints info logs to STDOUT not STDERR when initializing the enclave.
Because the JSON is also sent to STDOUT it results in invalid formatting when using a shell redirect.

Think these are the lines in question:

https://github.com/scrtlabs/SecretNetwork/blob/master/cosmwasm/packages/sgx-vm/src/enclave.rs#L106
https://github.com/scrtlabs/SecretNetwork/blob/master/cosmwasm/packages/sgx-vm/src/seed.rs#L103 (also line 112)

Sorry, I'd normally attach a PR for this because I know it's a super-simple fix, but I don't know Rust.

toml01:
Fixes #942


Cashmaney:
Isn't this already solved by https://github.com/scrtlabs/cosmos-sdk/pull/185?

toml01:
@Cashmaney it is solved, I wasn't aware because the issue was still open

But anyways, we have two different problems here.
1. The export thing itself (solved by @liorbond)
2. How do we want Rust logs to work? we can either have 1. all to stdout 2. all to stderr 3. error/warnings to stderr and info/trace to stdout.
I tend to vote for 2, because production builds only allows `warn` and `error` anyway, and it's easier to implement (@assafmo and @reuvenpo also thought 2 is better IMO).

As for the env var, I'm not sure how useful it is so I don't have a string opinion on that

liorbond:
@toml01 @Cashmaney 
Actually didn't think about the production that only supports warnings and errors, assuming we will keep doing that in production I tend to agree with @toml01.
As a general idea, I think that only warnings and errors should be printed to stderr so people can filter out their errors easily

toml01:
Keep in mind that splitting logs between `stdout` and `stderr` would require to impl. our own logger. Which is pretty easy, but just saying.
In general I think that the best practice is for info, traces, diagnostics, etc to go to stderr, and actual command output should go to stdout. In our case, everything should go to stderr IMO. Users can later filter based on the log level (and that's a good argument for having an env. var. controlled log level btw)

assafmo:
I'm with Tom. Cosmos/Tendermint logs should have consistent behavior with the rest of Cosmos, enclave logs should go to stderr.

assafmo:
What are we doing with this, and can we rebase onto `cosmwasm-v1`?

toml01:
You can rebase, I based it on master because it was not cosmwasm related.
@Cashmaney @liorbond final thoughts/disclaimers?

liorbond:
https://github.com/scrtlabs/SecretNetwork/pull/1134

PR #1092, opened 2022-08-24
Fix possible null-dref

liorbond:
`GetSignaturesV2` can return err which means `signatures` gonna be null now when drefing `signatures` in line 243: `signatures[pkIndex]` it might crash


toml01 (approved):

assafmo:
https://github.com/scrtlabs/SecretNetwork/commit/40c02fefe416cfd38b1a7e52b7c233892fbee657#diff-9c158daeb4dc0e28b2687e71df546511e741c2c29313cd2cb27595ea1ac73506R220-R223 :eyes:

Issue #754, opened 2021-05-26
Null response from `DistQuery::Rewards` when there are no rewards


baedrik:
When a secret contract attempts to do the Rewards query from the Distribution module like follows (where delegator is the HumanAddr whose rewards you want to view):
```rust
    let rewards_resp: RewardsResponse = deps.querier.query(&QueryRequest::Dist(DistQuery::Rewards {
       delegator,
    }))?;
```
If the delegator does not have any pending rewards, the query returns a null response which causes a parse error.  Instead it should always return a valid `RewardsResponse`, but when there are no rewards, the `rewards` and `total` fields should be empty Vecs.

assafmo:
@liorbond is this fixed in v1.4?

PR #1094, opened 2022-08-25
Fix v010 DistQuery::Rewards bug

eladr7:
When making a DistQuery::Rewards, where the delegator does not have any pending rewards the query returns null for rewards. e.g. {"rewards":null,"total":[]}
The desired result is to return an empty vector in such a case; e.g. {"rewards":[],"total":[]}

*I will change the fix to be more pretty in an upcoming update

### DIFF
--- a/go-cosmwasm/types/queries.go
+++ b/go-cosmwasm/types/queries.go
@@ -469,8 +469,8 @@ type RewardsQuery struct {
 
 // DelegationResponse is the expected response to DelegationsQuery
 type RewardsResponse struct {
-	Rewards []Rewards   `json:"rewards,omitempty"`
-	Total   RewardCoins `json:"total,omitempty"`
+	Rewards []Rewards   `json:"rewards"`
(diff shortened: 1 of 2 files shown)


assafmo:
Hi, please point the PR to the branch you came out of, rather than master

assafmo:
Also, without the ability to upgrade contracts I'm not sure I feel comfortable fixing this (i.e. breaking the current behavior), as we might already have contracts on mainnet that act upon "null as no rewards" and after this fix will stop functioning correctly.

We might want to wait for contract upgrades before we fix this, or at least reach out to all staking derivatives teams and ask whether this affects them.

Would love more feedback on this.

toml01:
#754

toml01:
Yes I tend to agree with @assafmo 
Looks like it got fixed on cosmwasm upstream. We should probably wait with this for contract upgrades or if we break backward compatibility in the future..

PR #1109, opened 2022-08-31
v1.4 upgrade handler

Cashmaney:


x/compute/module.go
```diff
@@ -100,10 +101,16 @@ func NewAppModule(keeper Keeper) AppModule {
 }
 
 // ConsensusVersion implements AppModule/ConsensusVersion.
-func (AppModule) ConsensusVersion() uint64 { return 1 }
+func (AppModule) ConsensusVersion() uint64 { return 2 }
```
assafmo, line 104: `ConsensusVersion()` needs to be incremented only if an existing data structure in the storage is changed in a non-backwards-compatible way, which AFIAK we don't have right now.

`ConsensusVersion()` is used only to decide whether to run the migration code while upgrading the chain, so it's unnecessary if our migration code is empty.

PR #1120, opened 2022-09-01
hermes docker waits for localsecrets

eshelB:
fixing ci issue where to relayer started up too early, where the connection between the ibc chains was not yet possible


assafmo:
Merged into #1121

PR #1124, opened 2022-09-04
make go tests start up faster when rerunning them

reuvenpo:
without this, rerunning go tests on my machine can take up to 30 seconds to start up


assafmo:
Can you rebase onto `cosmwasm-v1`?

reuvenpo:
@assafmo done

PR #1127, opened 2022-09-05
Fix EIP191 signature verification inside the enclave

assafmo:


cosmwasm/enclaves/shared/contract-engine/src/contract_validation.rs
```diff
@@ -343,6 +343,55 @@ fn get_signer_and_messages(
...
+                Some(start_index) => start_index,
+                None => {
+                    warn!(
+                        "SIGN_MODE_EIP_191 failed to find first occurance of '{{' in '{}'",
+                        sign_bytes_as_string
+                    );
+                    return Err(EnclaveError::FailedTxVerification);
+                }
+            };
+            let sign_doc_str = &sign_bytes_as_string[start_index..sign_bytes_as_string.len()];
```
Cashmaney, line 379: same as `let sign_doc_str = &sign_bytes_as_string[start_index..];`?

assafmo: Will try. Rust didn't let me for some reaosn.

PR #1133, opened 2022-09-06
Bump gas costs to 100K per contract instance and another 200_000 (was…

Cashmaney:
…m gas) per ocall (read/write from db)

### DIFF
--- a/x/compute/internal/keeper/keeper_test.go
+++ b/x/compute/internal/keeper/keeper_test.go
@@ -279,9 +279,10 @@ func TestInstantiate(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, "secret18vd8fpwxzck93qlwghaj6arh4p7c5n8978vsyg", contractAddr.String())
 
+	// gas can change +- 10% before we start failing, though maybe for consensus we should check a constant amount
 	gasAfter := ctx.GasMeter().GasConsumed()
-	require.Greater(t, gasAfter-gasBefore, uint64(10000))
-	require.Less(t, gasAfter-gasBefore, uint64(90000))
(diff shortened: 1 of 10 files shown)


liorbond (approved):

x/compute/internal/keeper/keeper_test.go
```diff
@@ -840,14 +841,6 @@ func TestExecuteWithStorageLoop(t *testing.T) {
 	ctx = ctx.WithGasMeter(sdk.NewGasMeter(gasLimit))
 	require.Equal(t, uint64(0), ctx.GasMeter().GasConsumed())
 
-	// ensure we get an out of gas panic
-	defer func() {
-		r := recover()
-		require.NotNil(t, r)
-		_, ok := r.(sdk.ErrorOutOfGas)
-		require.True(t, ok, "%+v", r)
-	}()
```
assafmo, line 849: Why delete this?

Cashmaney: Because the execution now returns a wasmerror gracefully rather than returning a panic

assafmo: I think the cosmos-sdk panics when out of gas, so when the execution returns the gas amount from WASM and we do `consumeGas(ctx, gasUsed)` it panics there... Am I wrong?

Cashmaney: This is because when calling `use_gas_externally` it doesn't actually consume gas on the sdk side (that would require another ocall), but just bumps up the gas tracker in the enclave. In this test now that gets hit, not the sdk gas, so the error is actually a graceful out of gas, rather than the panic on the sdk side

assafmo: I see, thanks for explaining

PR #1136, opened 2022-09-07
WIP: Cache enclave keys

Cashmaney:


cosmwasm/enclaves/shared/contract-engine/Cargo.toml
```diff
@@ -9,6 +9,7 @@ query-only = []
 debug-print = []
 test = []
 
+
```
liorbond, line 12: remove please :wink:

Cashmaney: No, this is critical

PR #1138, opened 2022-09-07
Implemented V1 tests for verifying Env(exec,qry) , MessageInfo(exec) …

liorbond:
…communicates well with Go. Same for V010(handle) and Env


assafmo:
Moved to #1162

PR #1165, opened 2022-09-11
USC module with `Enable=false`

Cashmaney:


app/upgrades/v1.4/upgrades.go
```diff
@@ -24,6 +26,26 @@ func CreateUpgradeHandler(mm *module.Manager, configurator module.Configurator,
 
 		ctx.Logger().Info("Running module migrations for v1.4...")
 
+		for moduleName := range mm.Modules {
+			vm[moduleName] = mm.Modules[moduleName].ConsensusVersion()
+		}
```
assafmo, line 31: Unnecessary

PR #1180, opened 2022-09-13
Custom log levels per sdk module workinggggg

Cashmaney:


Cashmaney:
No need for this - filter logs coming soonish

PR #1183, opened 2022-09-14
Ffi changes

reuvenpo:


Makefile
```diff
@@ -368,7 +368,7 @@ prep-go-tests: build-test-contract
 go-tests: build-test-contract
 	SGX_MODE=SW $(MAKE) build-linux-with-query
 	cp ./$(EXECUTE_ENCLAVE_PATH)/librust_cosmwasm_enclave.signed.so ./x/compute/internal/keeper
-	cp ./$(QUERY_ENCLAVE_PATH)/librust_cosmwasm_query_enclave.signed.so ./x/compute/internal/keeper
+	#cp ./$(QUERY_ENCLAVE_PATH)/librust_cosmwasm_query_enclave.signed.so ./x/compute/internal/keeper
```
assafmo, line 371: Why?

cosmwasm/enclaves/shared/contract-engine/src/db.rs
```diff
@@ -11,7 +11,7 @@ use crate::external::{ecalls, ocalls};
 use super::contract_validation::ContractKey;
 use super::errors::WasmEngineError;
 
-#[cfg(not(feature = "query-only"))]
+// #[cfg(not(feature = "query-only"))]
```
assafmo, line 14: Why all of these?

reuvenpo: To answer both your comments:

With this new design we don't need two enclaves because it supports both the query-only and regular workflows. For now I'm just commenting things out but imo we can just remove the query enclave completely

PR #1199, opened 2022-09-22
Fix secret message parsing when message is plaintext in IBC

liorbond:


assafmo (approved):
Tested with cw20-ics20 and it works :100:

PR #1217, opened 2022-10-19
LocalSecret: Fix future times

### LINKED ISSUE #1071: LocalSecret: lowering consensus / block time breaks block timestamp progression
# Issue: Time progression is messed up when speeding up the block time.

I'm currently using the official [block time speedup tutorial](https://docs.scrt.network/secret-network-documentation/development/tools-and-libraries/local-secret#pro-tip-speed-up-block-time).

On a remote localsecret instance I created on August 9th, `secretcli status` returns:

| Result | Correctness |
| --- | --- |
| `"earliest_block_time":"2022-08-09T07:31:18.30567358Z"` | ✅ |
| `"latest_block_time":"2022-08-28T21:29:55.862437404Z"` | ❌ |

It's August 13th today, so as you can see, the current block time is over 2 weeks in the future. I suspect this is because the legacy block time from the old dev container is hardcoded instead of pulling the latest timestamp every block. 

Please, please fix this when possible - being forced to use no block time speedup makes time-dependent contracts (vesting schedules, distributors, etc) a complete pain in the ass to reliably test and debug. 

**EDIT:** Alternatively, displayed block time could be a separately-configured option to either match the present time of block commit or be a predetermined, higher value (which would effectively allow for fast-tracking any contract test that requires waiting)

assafmo:
- Fix #1071
- Refactor changing default params in genesis.json


Cashmaney (dismissed):
This doesn't change all the referrences from stake to uscrt, so localsecret doesn't start properly

assafmo:
Ahh dammit, I couldn't debug because my laptop died. I wanted to refactor the genesis.json edits to be a bit more pleasant but we can revert that if it's a blocker

liorbond (approved):

assafmo:
Released as `localsecret:v1.6.0-patch.1`

PR #1233, opened 2022-10-26
Fix release

Cashmaney:
Fix release CI build


assafmo (approved):
Tag as v1.4.0-alpha.999 to be sure?

PR #1255, opened 2022-11-10
Benchmark tests: Added tests for viewing-key and permit

eladr7:


Cashmaney (changes requested):
Left a few comments. In general, it looks good, but please go over the comments I left. Also, instead of modifying the benchmarking contract it would have probably been less effort to just load the snip20 contract, but I guess it's okay to do it this way also. I guess now we can see if there's a significant difference in contract loading times with the toolkit and some heavier code included

x/compute/internal/keeper/bench_test.go
```diff
@@ -304,4 +341,129 @@ func TestRunBenchmarks(t *testing.T) {
...
+
+	// Turn permitString into bytes array
+	permitBytes := []byte(permitString)
+
+	// Convert permitBytes into a legacytx.StdSignDoc struct
+	var permitSignDoc legacytx.StdSignDoc
+	legacy.Cdc.MustUnmarshalJSON(permitBytes, &permitSignDoc)
+
+	// Convert permitSignDoc into bytes array again
+	permitSignDocBytes := sdk.MustSortJSON(legacy.Cdc.MustMarshalJSON(permitSignDoc))
```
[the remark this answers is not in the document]
eladr7: We don't have an example, now we have.
But I agree, maybe I should've generated a static permit and used it..

x/compute/internal/keeper/bench_test.go
```diff
@@ -304,4 +341,129 @@ func TestRunBenchmarks(t *testing.T) {
...
+	queryCases := map[string]struct {
+		bench            Bench
+		queryFunctionPtr func(
+			t *testing.T, keeper Keeper, ctx sdk.Context,
+			contractAddr sdk.AccAddress, creator sdk.AccAddress, creatorPriv crypto.PrivKey,
+			viewingKey string,
+		)
+		loops uint64
+	}{
+		"Query with viewing-key": {
```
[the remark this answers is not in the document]
eladr7: 100%

x/compute/internal/keeper/bench_test.go
```diff
@@ -304,4 +341,129 @@ func TestRunBenchmarks(t *testing.T) {
...
+			loops:            10,
+		},
+		"Query with permit": {
+			bench:            Balance,
+			queryFunctionPtr: PermitBenchRun,
+			loops:            10,
+		},
+	}
+
+	for name, tc := range queryCases {
```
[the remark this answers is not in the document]
eladr7: It was done in the execution stage.
Although, when I think of it, I should make sure set is coming after create, which is currently not enforced

x/compute/internal/keeper/bench_test.go
```diff
@@ -24,11 +32,27 @@ const (
 	BenchWriteStorage                   = "bench_write_storage"
 	BenchAllocate                       = "bench_allocate"
 	BenchReadLargeItemFromStorage       = "bench_read_large_item_from_storage"
+	CreateViewingKey                    = "create_viewing_key"
+	SetViewingKey                       = "set_viewing_key"
```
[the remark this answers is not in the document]
eladr7: It's more complex and that's why I thought it's worth benchmarking

cosmwasm/contracts/v1/compute-tests/bench-contract/src/contract.rs
```diff
@@ -44,17 +57,87 @@ pub fn execute(
...
+        _ => viewing_keys_queries(deps, msg),
+    }
 }
 
 #[entry_point]
 pub fn reply(_deps: DepsMut, _env: Env, _reply: Reply) -> StdResult<Response> {
     Ok(Response::default())
 }
+
+pub fn try_create_key(deps: DepsMut, env: Env, info: MessageInfo) -> StdResult<()> {
```
liorbond, line 83: This function shouldn't return anything and should be named "create_key" as the result is always Ok(()) in this case

cosmwasm/contracts/v1/compute-tests/bench-contract/src/contract.rs
```diff
@@ -44,17 +57,87 @@ pub fn execute(
...
+        deps.storage,
+        &info,
+        &env,
+        info.sender.as_str(),
+        "some_entropy".as_bytes(),
+    );
+    Ok(())
+}
+
+pub fn try_set_key(deps: DepsMut, info: MessageInfo, key: String) -> StdResult<()> {
```
liorbond, line 94: ditto

cosmwasm/contracts/v1/compute-tests/bench-contract/src/contract.rs
```diff
@@ -44,17 +57,87 @@ pub fn execute(
...
+    ViewingKey::set(deps.storage, info.sender.as_str(), key.as_str());
+    Ok(())
+}
+
+fn permit_queries(deps: Deps, permit: Permit, query: QueryWithPermit) -> Result<Binary, StdError> {
```
liorbond, line 99: In my opinion, the name is misleading here, I would suggest changing to the following:
```suggestion
fn query_with_permit(deps: Deps, permit: Permit, query: QueryWithPermit) -> Result<Binary, StdError> {
```

cosmwasm/contracts/v1/compute-tests/bench-contract/src/contract.rs
```diff
@@ -44,17 +57,87 @@ pub fn execute(
...
+            }
+
+            to_binary(&QueryAnswer::Balance {
+                amount: Uint128::from(BALANCE_QUERY_RESULT),
+            })
+        }
+    }
+}
+
+pub fn viewing_keys_queries(deps: Deps, msg: QueryMsg) -> StdResult<Binary> {
```
liorbond, line 128:
```suggestion
pub fn query_with_view_key(deps: Deps, msg: QueryMsg) -> StdResult<Binary> {
```

cosmwasm/contracts/v1/compute-tests/bench-contract/src/contract.rs
```diff
@@ -44,17 +57,87 @@ pub fn execute(
...
     Ok(Response::default())
 }
 
 #[entry_point]
-pub fn query(_deps: Deps, _env: Env, _msg: QueryMsg) -> StdResult<Binary> {
-    Ok(Binary::default())
+pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
+    match msg {
+        QueryMsg::WithPermit { permit, query } => permit_queries(deps, permit, query),
+        _ => viewing_keys_queries(deps, msg),
```
liorbond, line 74: Would add another type of QueryMsg dedicated for queries with viewing keys for further support.

cosmwasm/contracts/v1/compute-tests/bench-contract/src/msg.rs
```diff
@@ -27,8 +31,44 @@ pub enum ExecuteMsg {
...
+    CreateViewingKey {},
+    SetViewingKey {
+        key: String,
+        padding: Option<String>,
+    },
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum QueryWithPermit {
```
liorbond, line 43: Before reading the implementation of the enum I was sure the QueryWithPermit is a struct that contains a query and a permit. Would suggest to change here the name to "QueryType"

liorbond (changes requested):

x/compute/internal/keeper/bench_test.go
```diff
@@ -304,4 +341,129 @@ func TestRunBenchmarks(t *testing.T) {
...
+	expectedResult := `{"balance":{"amount":"42"}}`
+	queryAndVerify(t, keeper, ctx, contractAddr, vkQueryMsg, expectedResult)
+}
+
+func PermitBenchRun(
+	t *testing.T, keeper Keeper, ctx sdk.Context,
+	contractAddr sdk.AccAddress, creator sdk.AccAddress, creatorPriv crypto.PrivKey,
+	_ string,
+) {
+	permitSignature := generatePermitSignature(ctx, keeper, contractAddr, creator, creatorPriv)
```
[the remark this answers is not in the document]
Cashmaney: The changes make things a little better, but you're still benchmarking both createPermitQueryMsg which signs the permit and the query itself - I have a feeling that the 1-2 ms difference between permits and viewing keys might be because of it - signing something with a private key tends to be non-negligible when benchmarking

assafmo:
Any intermediate results to share? My hunch tells me permits are faster because VKs need to access the storage to get the viewing key.

Cashmaney:
> Any intermediate results to share? My hunch tells me permits are faster because VKs need to access the storage to get the viewing key.

These benchmarks run automatically via the Azure ci - you can go to the linked run and check it out there

Cashmaney (approved):

PR #1267, opened 2022-11-29
Added the tokenfactory module

eladr7:
The tokenfactory module allows a user to create a token of the notation: factory/{creator address}/{subdenom}


Cashmaney:
Closing this for now until we have more time to reprioritize

PR #1287, opened 2022-12-16
Fix OSX Ledger Disconnection Issues

chillyvee:
Fixes the most common ledger signing/disconnect errors with OS X + Ledger


Cashmaney:
Ty @chillyvee

assafmo:
Thanks @chillyvee!

PR #1295, opened 2023-01-03
Fixed parsing default type for CosmWasmMsg

Cashmaney:
Test available here: https://github.com/scrtlabs/secret.js/pull/135


assafmo:
Shouldn't we test it in https://github.com/scrtlabs/SecretNetwork/tree/master/integration-tests?

Cashmaney:
> Shouldn't we test it in https://github.com/scrtlabs/SecretNetwork/tree/master/integration-tests?

Added a test there too

PR #1298, opened 2023-01-05
Localsecret validator

eshelB:


assafmo:
AFAIK node_init.sh isn't used in LocalSecret

Cashmaney:
> AFAIK node_init.sh isn't used in LocalSecret

The PR name is wrong - there's no point in a single localsecret with this flag as it is a bootstrap anyway. It's for a network with multiple nodes that is used in a different image

Cashmaney (approved):

PR #1302, opened 2023-01-12
Match buildarg CHAINID in Makefile with Dockerfile

sbellem:
Unless I'm missing something the build arg in the Dockerfile is `CHAINID` meanwhile the Makefile uses `CHAIN_ID`.

This pull request changes the Makefile, meaning that `CHAINID` is chosen. If `CHAIN_ID` is preferred then I can change the Dockerfile instead. Just let me know.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -244,7 +244,7 @@ localsecret:
 			--build-arg SGX_MODE=SW \
 			$(DOCKER_BUILD_ARGS) \
  			--build-arg SECRET_NODE_TYPE=BOOTSTRAP \
- 			--build-arg CHAIN_ID=secretdev-1 \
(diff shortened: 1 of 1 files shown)


Cashmaney:
Hey @sbellem ! Thanks for the PR. I think we missed the difference in the argument names because a bit ago we moved the chain-id initialization to a run-time environment variable rather than a build time one (which you can find [here](https://github.com/scrtlabs/SecretNetwork/blob/master/deployment/docker/localsecret/bootstrap_init_no_stop.sh#L12)) so this variable gets ignored most of the time. Still, it's good to be consistent so I'll go ahead and merge this one

sbellem:
Thanks for the explanations @Cashmaney!

PR #1313, opened 2023-01-23
Bump cosmos-sdk version

Cashmaney:


assafmo:
Can you also rebuild swagger? (need to import the new protobufs into third_party/)

PR #1322, opened 2023-01-29
Ibc go v4

Cashmaney:


app/upgrades/v1.7/upgrade.go
```diff
@@ -22,7 +23,7 @@ const upgradeName = "v1.7"
 var Upgrade = upgrades.Upgrade{
 	UpgradeName:          upgradeName,
 	CreateUpgradeHandler: createUpgradeHandler,
-	StoreUpgrades:        store.StoreUpgrades{},
+	StoreUpgrades:        store.StoreUpgrades{Added: []string{icacontrollertypes.StoreKey}}, // we kind of forgot this in 1.3
```
assafmo, line 26: Not sure if this will work

assafmo (dismissed):
LGTM, just need to fix the secretjs test

assafmo:
I can also add packet forwarding tests later this week, shouldn't be hard

Cashmaney:
> I can also add packet forwarding tests later this week, shouldn't be hard

sababa, that's a different PR

assafmo:
This works but we still need to add the Fee & Packet Forwarding middlewares.

Cashmaney:
> There's a bug in our ibc-go integration where IBC proposals don't register with our CLI gov subcommand, e.g. `secretcli tx gov submit-proposal update-client` cannot be used. This was the case with v3 and also now with v4 so it's not an ibc-go bug. The proposal itself works (via secretjs - https://www.mintscan.io/secret/proposals/235) so it's just a CLI subcommand issue.

Is this super important for anything? If not, let's just open a bug and handle it outside the scope of this PR

assafmo:
> > There's a bug in our ibc-go integration where IBC proposals don't register with our CLI gov subcommand, e.g. `secretcli tx gov submit-proposal update-client` cannot be used. This was the case with v3 and also now with v4 so it's not an ibc-go bug. The proposal itself works (via secretjs - https://www.mintscan.io/secret/proposals/235) so it's just a CLI subcommand issue.
> 
> Is this super important for anything? If not, let's just open a bug and handle it outside the scope of this PR

Sorry, already fixed before I saw your comment. :grimacing:

PR #1326, opened 2023-01-31
Part of the DCAP rework: separate enclave networking into new crate -

liorbond:
* Removed ocalls that just open a socket directly
* Remove some unwraps() in favor of error handling to provide more clear errors


cosmwasm/enclaves/shared/net/src/consts.rs
```diff
@@ -0,0 +1,7 @@
+pub const SEED_SERVICE_DNS: &str = "sssd.scrtlabs.com";
+pub const DEFAULT_PORT: u16 = 443;
+
+// #[cfg(feature = "production")]
+// pub const SEED_SERVICE_DNS: &str = "sss.scrtlabs.com";
```
liorbond, line 5: why comment?

Cashmaney: Wasn't sure if we were going to end up using the production feature forever, and it didn't seem worth the trouble to pass it into the networking crate since this is just used for defaults anyway

liorbond: but you can't use the same dns name both for pulsar and for mainnet

Cashmaney: It matters not ser. You specify the dns when using the library (for example here: https://github.com/scrtlabs/SecretNetwork/pull/1326/files#diff-89fed89407503acd0b09cc73043fa216c4c6168ed79c26c4cdbe531f383b1889R624-R626)

PR #1329, opened 2023-02-06
Fix full sync in 1.7 - Split into convinient commits

liorbond:
When syncing a new node to 1.7 we will encounter an issue since we changed the format of seed.json
This PR solves the issue by creating a new cli command that the node operator will call in order to create old seeds


liorbond:
Still need to be covered: Auto-Register

deployment/docker/testnet/node_init.sh
```diff
@@ -69,6 +69,7 @@ then
   secretd q register secret-network-params 2> /dev/null
 
   secretd configure-secret node-master-key.txt "$SEED"
+  secretd configure-old-secret node-master-cert.der "$SEED"
```
Cashmaney, line 72: `configue-old-secret` is not a very intuitive name. Maybe configure-secret-legacy (or even a flag on the same configure-secret command?)

.gitignore
```diff
@@ -24,6 +24,8 @@ SHA256SUMS*
 *.der
 io-master-key.txt
 node-master-key.txt
+node-master-cert.der
```
liorbond, line 27:
```suggestion
```

.gitignore
```diff
@@ -24,6 +24,7 @@ SHA256SUMS*
 *.der
 io-master-key.txt
 node-master-key.txt
+io-master-cert.der
```
liorbond, line 27:
```suggestion
```

assafmo (approved):

PR #1331, opened 2023-02-08
Update encryption scheme for v1.7

assafmo:


cosmwasm/enclaves/shared/crypto/src/consts.rs
```diff
@@ -90,6 +90,8 @@ pub const CONSENSUS_CALLBACK_SECRET_DERIVE_ORDER: u32 = 4;
 
 pub const ENCRYPTED_KEY_MAGIC_BYTES: &[u8; 6] = b"secret";
 pub const CONSENSUS_SEED_VERSION: u16 = 2;
+/// STATE_ENCRYPTION_VERSION is bumped every time we change anything in the state encryption protocol
+pub const STATE_ENCRYPTION_VERSION: u32 = 3;
```
assafmo, line 94: Not sure if this is needed

toml01: Yes I think this is a good addition 👍🏼

cosmwasm/enclaves/shared/cosmwasm-types/v0.10/src/types.rs
```diff
@@ -119,6 +119,7 @@ pub struct TransactionInfo {
 pub struct BlockInfo {
     pub height: u64,
     /// Absolute time of the block creation in seconds since the UNIX epoch (00:00:00 on 1970-01-01 UTC).
+    /// Assaf: Since we added CosmWasm v1 this is in NANOseconds
```
toml01, line 122: Is it nanoseconds across the board now, or only in v1 contracts?

Cashmaney: The base type comes in nanos, then for v010 it's converted to seconds and v1 gets nanos + seconds

assafmo: In Tendermint it was always nanos, but in the beginning CosmWasm decided to pass it to the contract as seconds. Somewhere after v0.10 they decided to just pass it as is from Tendermint to the contract.

I added the comment here because on our implementation, the Go side passes everything to the enclave like it's a v0.10 contract (using the v0.10 types), and in the enclave if we detect that it's a v1 contract we convert the Env to its v1 types. When we added v1 we changed it so the Go side passes this field as nanos instead of seconds (we convert to seconds before passing it to a v0.10 contract), but the code comments inside the enclave still said it's in seconds.

toml01: Having both comments like that is pretty confusing 😅

assafmo: Will be fixed, thanks!

cosmwasm/enclaves/shared/cosmwasm-types/v0.10/src/types.rs
```diff
@@ -118,7 +118,9 @@ pub struct TransactionInfo {
 #[derive(Serialize, Deserialize, Clone, Default, Debug, PartialEq)]
 pub struct BlockInfo {
     pub height: u64,
-    /// Absolute time of the block creation in seconds since the UNIX epoch (00:00:00 on 1970-01-01 UTC).
+    /// Absolute time of the block creation in nanoseconds since the UNIX epoch (00:00:00 on 1970-01-01 UTC).
+    /// Note: when passed down to the enclave from Go this field is in nanoseconds, when passed down to a v0.10 contract this field is in seconds
+    /// For more context: https://github.com/scrtlabs/SecretNetwork/pull/1331#discussion_r1113526524
```
toml01, line 123: thanks ❤️

toml01 (approved):

PR #1344, opened 2023-02-21
Light client

Cashmaney:
This PR adds light-client validation to the enclave


cosmwasm/enclaves/shared/block-verifier/src/validator_whitelist.rs
```diff
@@ -0,0 +1,54 @@
...
+        self.0.len()
+    }
+
+    fn contains(&self, input: &String) -> bool {
+        self.0.contains(input)
+    }
+}
+
+pub fn whitelisted_validators_in_block(untrusted_block: &UntrustedBlockState) -> bool {
+    untrusted_block.validators.validators().iter().filter(|&a| {VALIDATOR_WHITELIST.contains(&a.address.to_string())}).count() >= VALIDATOR_THRESHOLD
```
eshelB, line 39: Could someone submit a block where a validator appears many times, which then causes one whitelisted validator to be able to pass the light client verification? maybe we need to do a unique()

assafmo: Interesting idea. Although This would change the block hash and therefore the signatures won't verify correctly.

assafmo (approved):

assafmo (dismissed):

cosmwasm/enclaves/shared/block-verifier/src/validator_whitelist.rs
toml01: @eshelB check this out: https://github.com/scrtlabs/tendermint-rs/blob/4019839ff894e531fbe4187f8461e4395d203220/light-client-verifier/src/operations/voting_power.rs#L128-L132

toml01 (approved):

PR #1408, opened 2023-04-19
IBC Hooks

assafmo:
- IBC Hooks
  - WASM Hooks - initiate a contract call with an incoming IBC token transfer.
  - WASM ack callbacks - get the IBC ack/timeout in non-IBC contracts that are using `IbcMsg::Transfer`.
  - See docs in https://github.com/scrtlabs/SecretNetwork/blob/ibc-hooks/x/ibc-hooks/README.md
- Added `memo` to `IbcMsg::Transfer` in https://github.com/scrtlabs/cosmwasm/pull/19. However devs can still use an older version of cosmwasm-std and it would still be compatible with [the added support for memo in x/compute](https://github.com/scrtlabs/SecretNetwork/pull/1408/commits/101f4f7edb41ae079e67bf721565dbf5f5cc0882).
- Added code comments about why plaintext replies (resulted from an IBC contract call) are not going through sender verification (basically because `msg.sender` isn't passed to `reply()`).

Tests: https://github.com/scrtlabs/secret.js/blob/ibc-go-v4/test/ibc.test.ts


x/ibc-hooks/wasm_hook.go
```diff
@@ -0,0 +1,488 @@
...
+		// If the ack is not a json object, error
+		return err
+	}
+
+	// Execute the contract
+	msg := []byte(fmt.Sprintf(
+		`{"ibc_lifecycle_complete":{"ibc_ack":{"channel":"%s","sequence":%d,"ack":%s,"success":%s}}}`,
+		packet.SourceChannel,
+		packet.Sequence,
+		ackAsJson,
```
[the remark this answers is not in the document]
assafmo: Cool

app/keepers/keepers.go
```diff
@@ -62,6 +63,10 @@ import (
 
 	ibcfeekeeper "github.com/cosmos/ibc-go/v4/modules/apps/29-fee/keeper"
 	ibcfeetypes "github.com/cosmos/ibc-go/v4/modules/apps/29-fee/types"
+
+	ibchooks "github.com/scrtlabs/SecretNetwork/x/ibc-hooks"
+	ibchookskeeper "github.com/scrtlabs/SecretNetwork/x/ibc-hooks/keeper"
```
eshelB, line 68: Did we customize the ibc-hooks module for our purposes? why didn't we import it from osmosis but brought all the files here instead?

eshelB (approved):

cosmwasm/enclaves/shared/cosmwasm-types/generic/src/lib.rs
```diff
@@ -73,15 +73,19 @@ impl BaseEnv {
...
+    pub fn into_versioned_env(
+        self,
+        api_version: &CosmWasmApiVersion,
+        is_empty_sender: bool,
+    ) -> CwEnv {
         match api_version {
-            CosmWasmApiVersion::V010 => self.into_v010(),
-            CosmWasmApiVersion::V1 => self.into_v1(),
+            CosmWasmApiVersion::V010 => self.into_v010(is_empty_sender),
+            CosmWasmApiVersion::V1 => self.into_v1(is_empty_sender),
```
Cashmaney, line 83: I'm not a big fan of complicating the interfaces with parameters like this. Why not just add a "set_sender" function and just call env.set_sender("")?

x/compute/internal/keeper/keeper.go
```diff
@@ -487,7 +487,7 @@ func (k Keeper) Instantiate(ctx sdk.Context, codeID uint64, creator sdk.AccAddre
 }
 
 // Execute executes the contract instance
-func (k Keeper) Execute(ctx sdk.Context, contractAddress sdk.AccAddress, caller sdk.AccAddress, msg []byte, coins sdk.Coins, callbackSig []byte) (*sdk.Result, error) {
+func (k Keeper) Execute(ctx sdk.Context, contractAddress sdk.AccAddress, caller sdk.AccAddress, msg []byte, coins sdk.Coins, callbackSig []byte, ibcHooks bool) (*sdk.Result, error) {
```
Cashmaney, line 490: Not a fan of adding a special parameter for ibcHooks. This bloats the interface even more than it is currently is. Isn't there any other way to do this?

cosmwasm/enclaves/shared/contract-engine/src/contract_operations.rs
```diff
@@ -275,20 +275,46 @@ pub fn handle(
...
+    // - Plaintext replies (resulting from an IBC call) - doesn't have a msg.sender anyway
+    // - IBC WASM Hooks
+    // - (In the future:) ICA
+    //
+    // We do want to allow executing contracts with plaintext input via IBC
+    // while the sender of an IBC packet cannot be verified.
+    // And we don't want malicious actors using this enclvae setting to just
+    // impersonate any sender they want.
+    // => Therefore we'll use a null sender if it cannot be verified && the input is plaintext
+    let is_null_sender = if should_validate_sig_info {
```
Cashmaney, line 289: If we already know we're running with ibchooks (from the keeper execute param), why do we need another check here? Couldn't we just add another handle type and avoid all the logic here?

Cashmaney (changes requested):

app/keepers/keepers.go
assafmo: Yeah we needed to change some code that was working with wasmd but not for us (privacy)

cosmwasm/enclaves/shared/contract-engine/src/contract_operations.rs
assafmo: Done in https://github.com/scrtlabs/SecretNetwork/commit/c11c681d1, LMK what you think

cosmwasm/enclaves/shared/cosmwasm-types/generic/src/lib.rs
assafmo: Reverted in 4b8ddafe382deef5faa305f289d48cfe46e068ed

cosmwasm/enclaves/shared/contract-engine/src/contract_operations.rs
```diff
@@ -275,20 +275,46 @@ pub fn handle(
...
+                    true // null sender
+                } else {
+                    return Err(err);
+                }
+            }
+            _ => false, // don't null sender
+        }
+    } else {
+        false // don't null sender
+    };
```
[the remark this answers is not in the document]
assafmo: Made it much simpler & readable in 4b8ddafe382deef5faa305f289d48cfe46e068ed

PR #1410, opened 2023-04-23
Flush the cache in a random order

liorbond:


Cashmaney:
Is this done yet?

liorbond:
> Is this done yet?

Practically, yes, fixing clippy and testing again
But I strongly suggest not adding it as part of v1.9!

Cashmaney:
> > Is this done yet?
> 
> Practically, yes, fixing clippy and testing again But I strongly suggest not adding it as part of v1.9!

I know, just want to make sure we have this ready so we don't have to context switch back to it later

PR #1412, opened 2023-04-24
Fix seed.json file to fit the old seed

liorbond:


Cashmaney (changes requested):
LGTM - just fix the tests

PR #1418, opened 2023-04-30
Ibc switch middleware

Cashmaney:


x/compute/internal/keeper/handler_plugin.go
```diff
@@ -113,6 +115,11 @@ func (h IBCRawPacketHandler) DispatchMsg(ctx sdk.Context, _ sdk.AccAddress, cont
 	if msg.IBC == nil || msg.IBC.SendPacket == nil {
 		return nil, nil, types.ErrUnknownMsg
 	}
+
+	if h.emergencyButton.IsHalted(ctx) {
+		return nil, nil, sdkerrors.Wrapf(types.ErrExecuteFailed, "ibc contract execution is disabled")
+	}
+
```
assafmo, line 122: Why is this needed here? I thought that this is covered by putting the emergency button middleware first in the compute stack.

Is there a test for this in the integration tests/secret.js?

Cashmaney: When sending from a contract the packet goes directly to the channelkeeper and doesn't hit the middleware's SendPacket method. I'm not 100% sure why, or what needs to get plugged in to what to work, but we can also fix it later. I'm also pretty sure for this reason the ibcfee middleware doesn't work properly

PR #1443, opened 2023-05-17
Fix ica controller initialization

Cashmaney:


app/app.go
```diff
@@ -308,6 +310,29 @@ func (app *SecretNetworkApp) Name() string { return app.BaseApp.Name() }
 
 // BeginBlocker application updates every begin block
 func (app *SecretNetworkApp) BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock) abci.ResponseBeginBlock {
+	// Fix v1.9 mishap
+	if ctx.BlockHeight() < 8861820 {
```
Pipello, line 314: nit: it might be cleaner to have this as `const` such as `BlockHeightThresholdFixV19Mishap`.

assafmo: Thanks! This was a one-off on mainnet and will not get into master or any future version.

PR #1452, opened 2023-05-29
Verify new message type2

Cashmaney:


cosmwasm/enclaves/shared/block-verifier/src/verify/registration.rs
```diff
@@ -0,0 +1,43 @@
...
+            Ok(ra_msg) => {
+                if ra_msg.certificate == certificate {
+                    return true;
+                }
+                error!("Error failed to validate registration message - 0x7535");
+                false
+            }
+            Err(e) => {
+                debug!("Error decoding registation protobuf: {}", e);
+                error!("Error decoding msg from block validator - 0xA0F2");
```
toml01, line 28: What's the deal with this hex `0xA0F2`?

cosmwasm/enclaves/shared/block-verifier/src/verify/registration.rs
```diff
@@ -0,0 +1,43 @@
+use crate::VERIFIED_MESSAGES;
+use log::{debug, error};
+use protobuf::Message;
+
+pub fn verify_reg_msg(certificate: &[u8]) -> bool {
+    let mut verified_msgs = VERIFIED_MESSAGES.lock().unwrap();
```
toml01, line 6: Not relevant specifically to this PR but just a thought, should we log a certain error in case this `lock().unwrap()` fails? Debugging this would be hell

cosmwasm/enclaves/execute/Cargo.toml
```diff
@@ -52,6 +52,7 @@ enclave_contract_engine = { path = "../shared/contract-engine" }
 enclave_crypto = { path = "../shared/crypto" }
 enclave_utils = { path = "../shared/utils" }
 enclave_cosmos_types = { path = "../shared/cosmos-types", optional = true }
+cosmos_proto = { path = "../shared/cosmos-proto" }
```
toml01, line 55: I think you ended up not using it directly in this crate?

toml01 (dismissed):
Left a few small comments but overall looks good.
We should implement the message type check though before merging to master

cosmwasm/enclaves/shared/block-verifier/src/verify/registration.rs
Cashmaney: I kind of like adding random error codes - makes finding the exact error easier in a month or two when someone randomly pastes their error. Also we can one day create a codex of error codes & their meaning/more info, etc.

cosmwasm/enclaves/shared/block-verifier/src/verify/registration.rs
```diff
@@ -0,0 +1,43 @@
...
+    let next = verified_msgs.get_next();
+
+    let result = if let Some(msg) = next {
+        // I assume @assaf will add the message type to the verified messages, so I'm going to leave this here for now
+        // without a check - it will fail in the parse either way
+
+        // if !check_message_is_reg(&msg) {
+        //     error!("Error failed to validate registration message - 0x7535");
+        //     false
+        // }
```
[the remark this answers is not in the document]
Cashmaney: Yeah I don't remember why I did this 🤣

cosmwasm/enclaves/shared/block-verifier/src/verify/registration.rs
Cashmaney: I actually think it would be kind of dramatic, as this would mean a failure in the lock or something? Not sure when this is even possible

assafmo: Yeah, it's going to give no details if it fails, but it should never fail.

PR #1471, opened 2023-06-29
secret-contract-optimizer: support multiple contracts

luca992:
Currently the `secret-contract-optimizer` only supports a single contract, not multiple contracts in a workspace. This updates the script in the Dockerfile to copy all built wasm files to a folder in the root of the project called `optimized-wasm` (naming up for debate). It then loops through them and runs `wasm-opt` each copied wasm. Then finally gzips them all and deletes them.

Also, I updated rust to 1.69.0... [rust 1.70.0 is not working](https://github.com/CosmWasm/cosmwasm/issues/1727)

### DIFF
--- a/deployment/dockerfiles/base-images/secret-contract-optimizer.Dockerfile
+++ b/deployment/dockerfiles/base-images/secret-contract-optimizer.Dockerfile
@@ -1,8 +1,15 @@
-FROM rust:1.68.0-slim-bullseye
+FROM rust:1.69.0-slim-bullseye
 
 RUN rustup target add wasm32-unknown-unknown
 RUN apt update && apt install -y binaryen clang && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /contract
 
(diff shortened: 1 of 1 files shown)


assafmo:
Hi @luca992, thanks for this! Can you provide an example of how to run this with multiple contracts?

luca992:
> Hi @luca992, thanks for this! Can you provide an example of how to run this with multiple contracts?

Yeah, I pushed it to dockerhub and for example https://github.com/eqoty-labs/snip721-migratable in the `Makefile`
It's the same command as normal.

```
.PHONY: compile-optimized-reproducible
compile-optimized-reproducible:
	docker run --rm -v "$$(pwd)":/contract \
		--mount type=volume,source="$$(basename "$$(pwd)")_cache",target=/code/target \
		--mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
		mr7uca/secret-contract-optimizer:0.0.1
```

PS. I didn't really realize that [rust-optimize](https://github.com/CosmWasm/rust-optimizer) should also work and it actively maintained. However, I found it doesn't work with any contract that uses `secret-toolkit-crypto` due to this [issue](https://github.com/CosmWasm/rust-optimizer/issues/126) with `secp256k1-sys`

assafmo (approved):

PR #1484, opened 2023-07-24
secret-contract-optimizer: only build libraries

kent-3:
If the contract contains a binary (such as this https://github.com/scrtlabs/secret-template/tree/master/src/bin) it will not work with the optimizer. The `--lib` flag should be included to only build the package's library.

Example error:
```
error: can't compile schema generator for the `wasm32` arch
       hint: are you trying to compile a smart contract without specifying `--lib`?
  --> src/bin/schema.rs:6:5
```


assafmo:
How does this work both ways?

kent-3:
> How does this work both ways?

I'm not sure what you mean. When using the optimizer you wouldn't want/need to build any binaries. But by specifying the `--lib` flag you can keep them in your src folder.

Right now if you want to use the optimizer, you need to remove any binaries from your src folder and Cargo.toml.

PR #1486, opened 2023-07-26
Block state cache

Cashmaney:


cosmwasm/enclaves/shared/contract-engine/src/contract_validation.rs
```diff
@@ -44,6 +44,12 @@ fn is_subslice(larger: &[u8], smaller: &[u8]) -> bool {
     false
 }
 
+#[cfg(feature = "light-client-validation")]
+pub fn is_last_msg_in_block() -> bool {
```
toml01, line 48: Is this here on purpose?

cosmwasm/enclaves/shared/contract-engine/src/wasm3/mod.rs
```diff
@@ -813,6 +818,24 @@ fn host_read_db(
...
+                        "read_db() error while trying to allocate {} bytes for the value",
+                        unwrapped.len(),
+                    );
+                    err
+                })?;
+
+            return Ok(ptr_to_region_in_wasm_vm as i32);
+        }
+    }
+
```
toml01, line 838: Wouldn't it be easier to have a single BLOCK_CACHE and read/write only from it? Seems like it just adds additional complexity to have both at the same time

Cashmaney: It's annoying to do, since you want to flush the cache to the state of the chain every msg (otherwise it gets messy if the cache is really large with gas and possible failures etc)

Doing it this way is a bit more complex logically, but simpler to keep track of what needs to get written to the cache

cosmwasm/enclaves/shared/block-verifier/src/ecalls.rs
```diff
@@ -245,10 +245,15 @@ pub unsafe extern "C" fn ecall_submit_block_signatures(
     {
         // AppHash calculation is different for the first block
         let rp = READ_PROOFER.lock().unwrap();
-        if header.height.value() != 1 && rp.app_hash != header.app_hash.as_bytes() {
+        if signed_header.header.height.value() != 1
```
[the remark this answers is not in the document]
Cashmaney: Same thing. It just saves cloning the header structure since it gets parsed into the signed_header

cosmwasm/enclaves/execute/src/sdk_entrypoints.rs
```diff
@@ -0,0 +1,59 @@
+use core::slice;
```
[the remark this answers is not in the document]
Cashmaney: Ny thinking was that logic that will never get reused anywhere makes sense to just put it under the main enclave since that should be easier to find

cosmwasm/enclaves/shared/contract-engine/src/contract_validation.rs
Cashmaney: Pretty sure I forgot about that 👍

cosmwasm/enclaves/execute/src/sdk_entrypoints.rs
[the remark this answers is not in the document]
toml01: I'm pretty sure we're violating this rule a lot 😅 but ok makes sense

PR #1498, opened 2023-08-03
SNIP-52: Add `env.transaction.hash`

assafmo:
Refs:
- https://github.com/SolarRepublic/SNIPs/blob/feat/snip-52/SNIP-52.md#notification-data-algorithms
- https://github.com/SolarRepublic/snip52-reference-impl
- https://demo.starshell.net

Cc @darwinzer0 @blake-regalia

Also fixed the test contracts to work with the new Rust memory optimizations.


cosmwasm/enclaves/shared/contract-engine/src/contract_validation.rs
```diff
@@ -859,7 +860,7 @@ fn verify_callback_sig_impl(
 
 #[allow(clippy::too_many_arguments)]
 fn verify_input_params(
-    sig_info: &SigInfo,
+    #[cfg(feature = "light-client-validation")] sig_info: &SigInfo,
```
toml01, line 863: Is this needed for the SNIP-52 stuff? If not I suggest adding it as a different PR

cosmwasm/enclaves/shared/cosmwasm-types/v0.10/src/types.rs
```diff
@@ -134,6 +134,11 @@ pub struct TransactionInfo {
     /// using the pair (`env.block.height`, `env.transaction.index`).
     ///
     pub index: u32,
+    /// The hash of the current transaction bytes.
+    /// aka txhash or transaction_id
+    /// hash = sha256(tx_bytes)
+    #[serde(default)]
+    pub hash: String,
```
toml01, line 141: Why are we adding it only to 0.10, and not v1 type?

cosmwasm/enclaves/shared/contract-engine/src/contract_validation.rs
assafmo: It's not, it's just to stop the compiler from complaining that this is not used when this flag is off.

cosmwasm/enclaves/shared/cosmwasm-types/v0.10/src/types.rs
toml01: Ahh right right, I know that but I just haven't noticed that v1's `Env` points to the same `TransactionInfo` as v0.10

toml01 (approved):

cosmwasm/enclaves/shared/contract-engine/src/contract_validation.rs
assafmo: https://github.com/scrtlabs/SecretNetwork/pull/1506

cosmwasm/enclaves/shared/contract-engine/src/contract_validation.rs
assafmo, line 1: Cc @toml01

PR #1505, opened 2023-08-08
Proof integration fix store

eshelB:
uses a version of the cosmos sdk that passes a multistore to queries which contains height-1 as well, not only the last height

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -685,8 +685,8 @@ github.com/ryanuber/columnize v0.0.0-20160712163229-9b3edd62028f/go.mod h1:sm1tb
 github.com/samuel/go-zookeeper v0.0.0-20190923202752-2cc03de413da/go.mod h1:gi+0XIa01GRL2eRQVjQkKGqKF3SF9vZR/HnPullcV2E=
 github.com/sasha-s/go-deadlock v0.3.1 h1:sqv7fDNShgjcaxkO0JNcOAlr8B9+cV5Ey/OB71efZx0=
 github.com/sasha-s/go-deadlock v0.3.1/go.mod h1:F73l+cr82YSh10GxyRI6qZiCgK64VaZjwesgfQ1/iLM=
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.19
 replace (
 	// dragonberry
 	github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
-	github.com/cosmos/cosmos-sdk => github.com/scrtlabs/cosmos-sdk v0.45.13-0.20230726163133-dc2b5eab1db5
+	github.com/cosmos/cosmos-sdk => github.com/scrtlabs/cosmos-sdk v0.45.13-0.20230808154448-1e7ad22811a4
```
assafmo, line 8: Seeing this https://github.com/scrtlabs/cosmos-sdk/commit/1e7ad22811a4, is there a way to pass `height-1` when we call this? I'm not sure what are the performance implications.

eshelB: If we pass height-1 it breaks all other queries other than wasm

eshelB: wasm doesn't care as it has the cache

eshelB: I told Itzik about this, we thought maybe we should see what happens and if there are performance implications, we know what to look for. Also memory implications, as now queries have the whole db. But probably none of these implications exist as these are all pointers to subsections of the iavl store?

toml01: If we see performance issues, we could probably even create a context with both `height` and `height - 1`

eshelB: I tried to find a way to do this but it seems that the multiStore there didn't support more than one specific version, or all of them

toml01 (changes requested):
This is a buggy cosmos-sdk version, causing `x-cosmos-block-height` (i.e. query at height) to stop working. We shouldn't merge this.

PR #1506, opened 2023-08-09
Stop the compiler from complaining that `sig_info` is not used

assafmo:
Stop the compiler from complaining that `sig_info` is not used in `verify_input_params()` when the `light-client-validation` feature is off


toml01 (approved):
💯

PR #1522, opened 2023-08-24
Hardcoded upgrade admins

eshelB:
probably this asks for a refactor to a new file for the hardcoded admins.
The addresses appear in the list in the proposal, but the hashes were queries from the chain. Let me know if you want to see the code that queried the hashes (built by chatGPT and tweaked by me).


assafmo (dismissed):

Cashmaney (changes requested):
Need to move all these to a separate file - it's too much bloat.

Also, it doesn't compile

Cashmaney (approved):

PR #1527, opened 2023-08-29
Add cli for migration

liorbond:


assafmo (commented):
Thanks, I forgot about that!

PR #1541, opened 2023-09-20
added another test where the submsg issues a failing bank message

eshelB:
Another case I wanted to know and specifically verify - whether contracts preserve their state even if the submessage triggers a message that goes to other modules which fails the entire transaction.

Case 1: contract creates submessage, submessage creates failing bank message. everything reverts. 
Case 2: contract creates 1. submessage that succeeds, 2. bank message that fails. everything reverts.
Case 3: contract creates submessage that suceeds. On reply endpoint, sends failing bank message. everything reverts.

Added comments explaning all tests up to now.

NOTE: I Observed a case where I think we differ from cosmwasm's documentation:
It states [here](https://github.com/CosmWasm/cosmwasm/blob/main/SEMANTICS.md#order-and-rollback) that:
> "Submessages (and their replies) are all executed before any messages. They also follow the depth first rules as with messages. Here is a simple example. Contract A returns submessages S1 and S2, and message M1. Submessage S1 returns message N1. The order will be: S1, N1, reply(S1), S2, reply(S2), M1"

I don't know whether they mean that regular messages are postponed to the end of the whole message chain, or whether they are postponed just after all the submessages created by the current submsg. (probably latter as that would make executions not have to be context-aware).
However, I think (based on a look into the code and the results of the tests) that regular messages are NOT postponed to the end of either. So the order will be based on the order of the addition to the result. See [here](https://github.com/scrtlabs/SecretNetwork/pull/1541/files#diff-e109a1ee312d5edadd84ff4f57453dabc7ae3ba306c23dcc8aac44b7ed706253R681).

This difference in outcomes is small when sdk message fails, because either way, the whole tx is reverted, but may
be present in cases of contract-queries-to-chain if the sdk message succeeds.


cosmwasm/contracts/v1/compute-tests/test-compute-contract/src/contract.rs
```diff
@@ -648,11 +645,44 @@ pub fn execute(deps: DepsMut, env: Env, info: MessageInfo, msg: ExecuteMsg) -> S
...
+
+            response = response.add_message(
+                CosmosMsg::Bank(BankMsg::Send {
+                    to_address: "non-existent".to_string(),
+                    amount: vec![Coin::new(100, "non-existent")],
+                })
+            );
+
+            // NOTE that if the submsg is added after the message, then it is processed in THAT order,
+            // which is different from what is stated in cosmwasm's docs here: https://github.com/CosmWasm/cosmwasm/blob/main/SEMANTICS.md#order-and-rollback
```
eshelB, line 682: Interesting note here that I think describes a change in our behaviour compared to cosmwasm's docs!
@toml01
@liorbond

cosmwasm/contracts/v1/compute-tests/test-compute-contract/src/contract.rs
```diff
@@ -2332,46 +2444,36 @@ pub fn reply(deps: DepsMut, env: Env, reply: Reply) -> StdResult<Response> {
         }
         (11338, SubMsgResult::Ok(SubMsgResponse { data, .. })) => {
             let   contract_addr=  match from_binary(&data.unwrap()) {
-            Ok(ExecuteMsg::SendMsgClearAdmin {
```
eshelB, line 2335: these are just indentation changes that I was triggered by

PR #1582, opened 2023-11-09
Action to automatically close stale issues and PRs

toml01:


assafmo:
Ohhhh I hate it when this bot closes my stale issues on other repos 😁

toml01:
@assafmo 😂 we have too many old ones in this repo IMO. We need to clear those out somehow - a lot of those are probably irrelevant by now or just not urgent enough.

assafmo:
I agree, even though there are only 26 open issues. We can probably move what's relevant manually to Monday and close the rest. We don't get a lot of new issues on a regular basis so I think a bot is unnecessary, but ultimately I don't really care if you merge this.

assafmo (approved):

PR #1599, opened 2023-12-26
Update hardcode-admins-on-v1.10.md

PrivatePixels:
Key not owned by individual


assafmo:
This has already passed an on-chain proposal, I don't see a reason to change it now

PrivatePixels:
> This has already passed as an on-chain proposal, I don't see a reason to change it now

Yes. Just clarifying that the key is not owned by an individual.

PR #1600, opened 2023-12-26
Fix Swagger/OpenAPI Schema

kent-3:
This pull request addresses two issues in the `swagger_legacy.yaml` file:

- Corrected `tyDeprecated:` typos to `type:`
- Removed redundant object definition in the `/reg/consensus-io-exch-pubkey` endpoint response schema

Though part of the deprecated API, these errors were causing issues in the generated `openapi.yaml` and `swagger.yaml` files.

With these changes, the spec will now pass validation used by the [OpenAPI Generator tool](https://github.com/OpenAPITools/openapi-generator) when generating code.

### DIFF
--- a/client/docs/static/openapi/openapi.yaml
+++ b/client/docs/static/openapi/openapi.yaml
@@ -38971,7 +38971,7 @@ paths:
                         type: string
                         example: "26354"
                       tags:
-                        tyDeprecated: array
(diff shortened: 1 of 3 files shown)


assafmo:
Thanks! Looks good, I just want to see the CI building this before we merge

PR #1629, opened 2024-03-07
Dcap 04

valdok:


cosmwasm/enclaves/execute/src/registration/onchain.rs
```diff
@@ -7,19 +7,83 @@ use std::panic;
...
+}
+
+#[cfg(not(feature = "light-client-validation"))]
+fn get_current_block_time_s() -> i64
+{
+    return 0 as i64;
+}
+
+
+pub fn split_combined_cert(
```
Cashmaney, line 51: Can this function be private? Also why not just pass the cert_slice in here? You should avoid using raw pointers wherever possible

cosmwasm/enclaves/execute/src/registration/onchain.rs
```diff
@@ -7,19 +7,83 @@ use std::panic;
...
+        {
+            vec_cert = unsafe { slice::from_raw_parts(cert.offset(n0 as isize), s0 as usize).to_vec() };
+            vec_quote = unsafe { slice::from_raw_parts(cert.offset((n0 + s0) as isize), s1 as usize).to_vec() };
+            vec_coll = unsafe { slice::from_raw_parts(cert.offset((n0 + s0 + s1) as isize), s2 as usize).to_vec() };
+        }
+
+    }
+
+    (vec_cert, vec_quote, vec_coll)
+}
```
Cashmaney, line 85: Try to avoid using unsafe - by doing this you're losing every guarantee that Rust provides. Unless you're doing FFI or something you should be using safe functions everywhere. Also makes your code way, way more readable. Something like (I just chatgpt'd it, don't take this as gospel):

```rust
fn split_combined_cert(cert_slice: &[u8]) -> (Vec<u8>, Vec<u8>, Vec<u8>) {
    let mut vec_cert: Vec<u8> = Vec::new();
    let mut vec_quote: Vec<u8> = Vec::new();
    let mut vec_coll: Vec<u8> = Vec::new();

    // This assumes the format of the cert_slice starts with three u32 integers indicating the length of each part.
    let header_size = std::mem::size_of::<u32>() * 3;

    if cert_slice.len() >= header_size {
        // Attempt to read the sizes from the slice.
        // We're using byteorder crate for more readable handling of byte order conversions.
        // Ensure you have byteorder in your Cargo.toml!
        // byteorder = "*"
        use byteorder::{ByteOrder, LittleEndian};

        let s0 = LittleEndian::read_u32(&cert_slice[0..4]) as usize;
        let s1 = LittleEndian::read_u32(&cert_slice[4..8]) as usize;
        let s2 = LittleEndian::read_u32(&cert_slice[8..12]) as usize;

        // Verify we have enough data in the slice to cover these sizes
        if cert_slice.len() >= header_size + s0 + s1 + s2 {
            let base = header_size;
            vec_cert = cert_slice[base..base + s0].to_vec();
            vec_quote = cert_slice[base + s0..base + s0 + s1].to_vec();
            vec_coll = cert_slice[base + s0 + s1..base + s0 + s1 + s2].to_vec();
        }
    }

    (vec_cert, vec_quote, vec_coll)
}
```

cosmwasm/enclaves/execute/src/registration/onchain.rs
```diff
@@ -56,21 +120,83 @@ pub unsafe extern "C" fn ecall_authenticate_new_node(
         return NodeAuthResult::SignatureInvalid;
     }
 
-    let result = panic::catch_unwind(|| -> Result<Vec<u8>, NodeAuthResult> {
+    let mut target_public_key: [u8; 32] = [0u8; 32];
+
+    let (vec_cert, vec_quote, vec_coll) = split_combined_cert(cert, cert_len);
+
+    if vec_quote.is_empty() || vec_coll.is_empty() {
```
Cashmaney, line 127: This seems like a hacky way to tell whether we're validating for epid or dcap - can't we just add a field here to the function call with the type of attestation?

cosmwasm/enclaves/execute/src/registration/onchain.rs
```diff
@@ -56,21 +120,83 @@ pub unsafe extern "C" fn ecall_authenticate_new_node(
...
+                return NodeAuthResult::InvalidCert;
+            }
+        };
+
+        let res2 = verify_ra_report(report_body.mr_signer.m, report_body.mr_enclave.m, Some(SigningMethod::MRSIGNER));
+        if NodeAuthResult::Success != res2 {
+            return res2;
+        }
+
+        target_public_key.copy_from_slice(&report_body.report_data.d[..32]);
```
Cashmaney, line 195: Should split this into a separate function (and the epid logic as well)

cosmwasm/enclaves/execute/src/registration/onchain.rs
```diff
@@ -56,21 +120,83 @@ pub unsafe extern "C" fn ecall_authenticate_new_node(
...
+                }
+                r.0
+            }
+            Err(e) => {
+                trace!("Remote quote verification failed: {}", e);
+                return NodeAuthResult::InvalidCert;
+            }
+        };
+
+        let res2 = verify_ra_report(report_body.mr_signer.m, report_body.mr_enclave.m, Some(SigningMethod::MRSIGNER));
```
Cashmaney, line 190: res2 could probably be `verification_result` or something more informative

cosmwasm/enclaves/execute/src/registration/onchain.rs
```diff
@@ -56,21 +120,83 @@ pub unsafe extern "C" fn ecall_authenticate_new_node(
...
+                if r.1 != sgx_ql_qv_result_t::SGX_QL_QV_RESULT_OK {
+                    trace!("WARNING: {}", r.1);
+                }
+                r.0
+            }
+            Err(e) => {
+                trace!("Remote quote verification failed: {}", e);
+                return NodeAuthResult::InvalidCert;
+            }
+        };
```
Cashmaney, line 188: Question about this - if dcap isn't properly installed will this fail? Do we have any way to guarantee that if the node starts there aren't any driver dependencies here that will cause an error? I'm asking because the results of this function have to be the same for each and every node. Anyone that doesn't conform with the consensus will crash (or even halt the chain if there are too many)

cosmwasm/enclaves/execute/src/registration/attestation.rs
```diff
@@ -289,6 +292,230 @@ pub fn get_mr_enclave() -> [u8; 32] {
...
+            &mut my_report
+        )
+    };
+
+    if res != sgx_status_t::SGX_SUCCESS {
+        trace!("sgx_create_report = {}", res);
+    }
+
+    let mut vec_quote : Vec<u8> = Vec::new();
+    vec_quote.resize(quote_size as usize, 0);
```
Cashmaney, line 437:
```rust
let vec_quote: Vec<u8> = vec![0; quote_size as usize];
```

cosmwasm/enclaves/execute/src/registration/attestation.rs
```diff
@@ -289,6 +292,230 @@ pub fn get_mr_enclave() -> [u8; 32] {
...
+    if res != sgx_status_t::SGX_SUCCESS {
+        trace!("res = {}", res);
+    }
+
+    if rt != sgx_status_t::SGX_SUCCESS {
+        trace!("rt = {}", rt);
+    }
+
+    let mut vec_coll : Vec<u8> = Vec::new();
+    vec_coll.resize(0x4000, 0);
```
Cashmaney, line 457:
```rust
let vec_quote: Vec<u8> = vec![0; 0x4000 as usize];
```

Also why 0x4000? Avoid using magic numbers where you can

cosmwasm/enclaves/execute/src/registration/attestation.rs
```diff
@@ -289,6 +292,230 @@ pub fn get_mr_enclave() -> [u8; 32] {
...
+        vec_quote.as_ptr(),
+        vec_quote.len() as u32,
+        &qe_report,
+        exp_time_s,
+        exp_status,
+        qv_result,
+        p_supp.as_ptr(),
+        n_supp,
+        qve_isvsvn_threshold) };
+
```
Cashmaney, line 366: Does this verification happen inside an enclave? Where do we cryptographically verify the attestation? i.e. how do we know someone isn't just faking a response

cosmwasm/enclaves/execute/src/registration/attestation.rs
```diff
@@ -289,6 +292,230 @@ pub fn get_mr_enclave() -> [u8; 32] {
     rsgx_self_report().body.mr_enclave.m
 }
 
+
+#[cfg(feature = "SGX_MODE_HW")]
+pub fn verify_quote_ecdsa(
```
Cashmaney, line 297: Needs some description on the logic that's happening here (and in other complex functions)

cosmwasm/enclaves/execute/src/registration/attestation.rs
```diff
@@ -289,6 +292,230 @@ pub fn get_mr_enclave() -> [u8; 32] {
...
+
+    trace!("ECDSA quote size = {}", quote_size);
+
+    let mut report_data: sgx_report_data_t = sgx_report_data_t::default();
+    let mut my_report: sgx_report_t = sgx_report_t::default();
+
+    report_data.d[..32].copy_from_slice(pub_k);
+
+    res = unsafe {
+        sgx_create_report(
```
Cashmaney, line 425: Why use sgx_create_report and not rsgx_create_report?

Cashmaney (dismissed):
Okay, it's quite a bit of code but I think I reviewed most of it. I focused mainly on the stuff inside the enclave since that should be most of the critical path. Either way, I think these comments are a good place to start. 

For the most part it looks good - I think especially for someone that hasn't had much experience with Rust it's in a pretty good state.

Some general stuff:

I suggest splitting all the dcap and EPID stuff into different files - it's not entirely clear where one ends and where another begins, and so it's hard to see whether all the code paths that currently rely on epid have been ported to dcap. 

We could also use a bit more comments on what exactly is going on here, as all the validation stuff isn't trivial. 

Try to use more informative naming where possible (`size_dcap_q` or `size_dcap_s` are not great)

If you're using unsafe anywhere you should comment on why you're doing that, and why it's okay - if you're not specifically doing FFI a lot of the times you might find that you're doing something unnecessary

cosmwasm/enclaves/execute/src/registration/attestation.rs
valdok: In DCAP this is kinda complicated.

The verification consist of bringing all the relevant info (the original enclave quote, the so-called collateral, and the current time), and then verifying all this at-once.
The actual verification is performed by yet another enclave, provided by intel, called Quote Verificaiotn Enclave, QvE.

Here specifically the code that runs within our enclave. First it calls ocall_verify_quote_ecdsa() to trigger our code outside the enclave to check the attestation. It would load a QvE, provide it all the given information, the QvE would check all this and return the results, SIGNED by it. Those signed results are returned from ocall_verify_quote_ecdsa() here.

Then we call sgx_tvl_verify_qve_report_and_identity(), that is supposed to verify that the returned results are indeed signed by the alleged QvE (i.e. not forged or tempered with by the untrusted code).

cosmwasm/enclaves/execute/src/registration/attestation.rs
valdok: That's a "reasonable" buffer size for the expected result.
If the result won't fit - it'll be truncated, then the code in the enclave would reallocate the needed buffer size and call again.
Using a sane buffer size would spare this redundant call.

cosmwasm/enclaves/execute/src/registration/attestation.rs
```diff
@@ -289,6 +292,230 @@ pub fn get_mr_enclave() -> [u8; 32] {
...
+        unsafe {
+            ocall_get_quote_ecdsa_collateral(
+                &mut rt as *mut sgx_status_t,
+                vec_quote.as_ptr(),
+                vec_quote.len() as u32,
+                vec_coll.as_mut_ptr(),
+                vec_coll.len() as u32,
+                &mut size_coll)
+        };
+    }
```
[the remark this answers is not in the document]
valdok: please see upper comment

cosmwasm/enclaves/execute/src/registration/attestation.rs
Cashmaney: In this case I would suggest just allocating double the space and calling once - I don't think it's worth the complexity and/or overhead in doing 2 ocalls. The alloc should be fairly cheap comparatively.

cosmwasm/enclaves/execute/src/registration/attestation.rs
Cashmaney: So `sgx_tvl_verify_qve_report_and_identity` calls the QvE directly? How do we know that the QvE is the one that performed the operation? I mean how do we know that someone didn't just hijack the function call and return a fake result?

cosmwasm/enclaves/execute/src/registration/attestation.rs
valdok: The size of collateral is not known in advance. Typically it's ~12K, but theoretically can be arbitrary large

cosmwasm/enclaves/execute/src/registration/onchain.rs
valdok: Theoretically this should always succeed (but need to check). This method does't demand a driver.

cosmwasm/enclaves/execute/src/registration/onchain.rs
valdok: I would like to refrain from changing the protocol (protobufs and etc.)
The argument "cert" is a single blob with 3 things serialized: epid certificate, dcap quote, and dcap collateral.

PR #1717, opened 2025-06-17
fix: remove duplicate logic

largemouth:
Both branches in if statement have same body. So remove duplicate logic to make code more clear.


largemouth:
friendly ping

